PR #4, opened 2015-08-07
Example of using Avro serializer and Avro generated classes in consumer-producer workflow

gwenshap:


gwenshap:
Ick. Accidentally created a "dirty" branch. Let me close and retry.

PR #3, opened 2015-08-07
Example of using Avro serializer with generated (specific) Avro objects

gwenshap:
It is super similar to what already exists in producer example, but using generated Avro objects.

PR #5, opened 2015-08-07
Consumer-Producer example using generated Avro classes

gwenshap:
This logically depends on PR #3, so please merge that one first


gwenshap:
pinging for review

specific-avro-consumer/README.md
```diff
@@ -0,0 +1,50 @@
+This example reads Avro records containing click information, as produced by AvroProducerExample.
```
ewencp: Maybe include the directory name for `AvroProducerExample` as well?

specific-avro-consumer/README.md
```diff
@@ -0,0 +1,50 @@
...
+We enrich them with session information and publish both to stdout and to another Kafka topic.
```
ewencp: Maybe show an example and explain what "session" means in this context. It makes sense to people used to web app data, but not everybody works on that type of data. Maybe just show pre-enriched and post-enriched, saying how sessions are inferred so they cannot be recorded when the original logs were created, resulting in the `null` entry. This application is processing the events, tracking the last event found for each user and their last page view, then treating any page view greater than 30 minutes later as a new session.

Additionally, might be worth mentioning that this works because of semantic partitioning on keys. Doesn't need a complete explanation, but a mention + link to docs might be nice.

specific-avro-consumer/README.md
```diff
@@ -0,0 +1,50 @@
...
+Also, this example works with data generated by AvroClicksProducer, so start by running that project first.
+    
+Then create a topic called sessions:
+
+    # Create page_visits topic
+    $ bin/kafka-topics.sh --create --zookeeper localhost:2181 --replication-factor 1 \
+      --partitions 1 --topic sessions
+      
+
+Then run the producer to produce 100 clicks:
```
ewencp: I think this text is left over from copy-paste, it's a bit confusing in this context since this app is not driven by a total number of clicks.

specific-avro-consumer/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+    $ java -cp target/uber-clickstream-sessionizing-consumer-1.0-SNAPSHOT.jar io.confluent.examples.consumer.AvroClicksSessionizer http://localhost:8081
+    
+You can validate the result by using the avro console consumer (part of the schema repository):
+
+    $ bin/kafka-avro-console-consumer --zookeeper localhost:2181 --topic sessions --from-beginning
```
ewencp: One thing I noticed is that all the output is `sessionId == 0`, which doesn't really help demonstrate (or even exercise!) the code. It looks like you need > 256 log records for the producer to make it reuse IP addresses and we need events separated by at least 30 minutes. The first one is easy to fix, the second is a bit harder since we just use the current time. Ideas to make the input a bit more interesting so we can actually see this code doing the right thing wrt detecting new sessions even if the user just runs these quickstart commands?

specific-avro-consumer/src/main/java/io/confluent/examples/consumer/AvroClicksSessionizer.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+public class AvroClicksSessionizer {
+    private final ConsumerConnector consumer;
+    private final KafkaProducer<String, LogLine> producer;
+    private final String inputTopic;
+    private final String outputTopic;
+    private String zookeeper;
```
ewencp: These can be `final` since they are all only set in the constructor.

specific-avro-consumer/src/main/java/io/confluent/examples/consumer/AvroClicksSessionizer.java
```diff
@@ -0,0 +1,140 @@
...
+        // currently hardcoding a lot of parameters, for simplicity
+        String zookeeper = "localhost:2181";
+        String groupId = "AvroClicksSessionizer";
+        String inputTopic = "clicks";
+        String outputTopic = "sessionized_clicks";
+        String url = args[0];
+
+        AvroClicksSessionizer sessionizer = new AvroClicksSessionizer(zookeeper, groupId, inputTopic, outputTopic, url);
+        sessionizer.run();
+
```
ewencp: nit: would be nice to clean up the extra whitespace

specific-avro-consumer/src/main/java/io/confluent/examples/consumer/AvroClicksSessionizer.java
```diff
@@ -0,0 +1,140 @@
...
+        this.inputTopic = inputTopic;
+        this.outputTopic = outputTopic;
+        this.url = url;
+    }
+
+    private Properties createConsumerConfig(String zookeeper, String groupId, String url) {
+        Properties props = new Properties();
+        props.put("zookeeper.connect", zookeeper);
+        props.put("group.id", groupId);
+        props.put("auto.commit.enable", "false");
```
ewencp: Might want to add a comment explaining why you're doing this. I'm assuming it is because for this simple demo it ensures you can re-run and still get useful output, but since `commit()` is never called anywhere it might need some additional explanation.

specific-avro-consumer/src/main/java/io/confluent/examples/consumer/AvroClicksSessionizer.java
```diff
@@ -0,0 +1,140 @@
...
+        props.put("schema.registry.url", url);
+        props.put("specific.avro.reader", true);
+
+        return props;
+    }
+
+    private void run() {
+        Map<String, Integer> topicCountMap = new HashMap<String, Integer>();
+        // Hard coding single threaded consumer
+        topicCountMap.put(inputTopic, new Integer(1));
```
ewencp: Can just use `1` instead of `new Integer(1)` here.

specific-avro-consumer/src/main/java/io/confluent/examples/consumer/AvroClicksSessionizer.java
```diff
@@ -0,0 +1,140 @@
...
+            }
+            event.setSessionid(sessionId);
+            System.out.println(event.toString());
+            ProducerRecord<String, LogLine> record = new ProducerRecord<String, LogLine>(outputTopic, event.getIp().toString(), event);
+            producer.send(record);
+        }
+
+    }
+
+    private KafkaProducer<String, LogLine> getProducer(String topic, String url) {
```
ewencp: topic parameter is not used in this method

ewencp:
@gwenshap Bunch of minor comments, but they're just that: minor. Other item I noticed that didn't fit in anywhere for inline comments is that there is no shutdown hook, which means both consumer and producer don't get to go through clean shutdown. The consumer could definitely use it since it cleanly shuts down each of the queues for the `KafkaStreams`; the producer can't do much useful until `flush()` is available.

specific-avro-consumer/README.md
gwenshap: Is there a specific place in the docs you think will match here?

gwenshap: "how sessions are inferred so they cannot be recorded when the original logs were created, resulting in the null entry"

This is where the example kind of breaks down. The "original app" could have the same in-memory map of visits that our consumer-producer has, so it could populate this field. I'm going to avoid the whole minefield - its just an example of a common pattern :)

specific-avro-consumer/README.md
gwenshap: I'm thinking of modifying the click generator to produce infinite clicks but sleep $N seconds between clicks. 

Then we can drop the session length to something completely unrealistic (10 sec? 30 sec?) for demo purposes.

Does that make sense? 
Batch demos are much easier since you can pre-cook the data, but I think current time is actually a nice feature.

specific-avro-consumer/README.md
ewencp: Hmm, seems nothing convenient wrt semantic partitioning is directly linkable. Maybe http://kafka.apache.org/documentation.html#theproducer ?

wrt the "original app", if it's a web app run behind a load balancer, plenty of people don't bother with sticky assignment to servers (or use LBs that don't implement it), so this pipeline is quite reasonable. Would be better if we actually persisted the data & did offset commit, but this works as a basic example.

specific-avro-consumer/README.md
ewencp: Yes, that makes sense. Something like 5 seconds is probably good enough. I don't want to get impatient waiting for simple demo code to run :)

gwenshap:
pinging for review :)

specific-avro-consumer/README.md
```diff
@@ -0,0 +1,70 @@
+This example reads Avro records containing click information, as produced by AvroProducerExample .
```
ewencp: Period and comma on this first line seem mixed up.

specific-avro-consumer/README.md
```diff
@@ -0,0 +1,70 @@
...
+(you can find the AvroProducerExample in specific-avro-producer directory),
+We then enrich these events with session information and publish both to stdout and to another Kafka topic.
+
+Sessionization is a very common step in preparing user click information collected from website logs for analysis. 
+It involves grouping user clicks into visits and marking all the clicks that belong to the same visit with unique session identifier.
+These sessions are later used to analyse what users are doing on the website during a visit, and answer questions such as "how many visits end in a purchase" or "what is the last page visited before purchasing".
+The way we decide if a visit from a user is a new session is by examining the time gap between his current and previous visit.
```
ewencp: I am not picky about this and we have no policy, but for pronouns I generally try to go by:

1. Neutral via [singular they](https://en.wikipedia.org/wiki/Singular_they) - our users are both genders!
2. Feminine - clearly specifies one gender, but generally accepted to cover both and simpler than option 3
3. The incredibly awkward feminine/masculine combo
4. Masculine

specific-avro-consumer/src/main/java/io/confluent/examples/consumer/AvroClicksSessionizer.java
```diff
@@ -0,0 +1,152 @@
...
+        KafkaStream stream = consumer.createMessageStreams(topicCountMap, stringDecoder, avroDecoder).get(inputTopic).get(0);
+        ConsumerIterator it = stream.iterator();
+        System.out.println("Ready to start iterating wih properties: " + props.toString());
+        System.out.println("Reading topic:" + inputTopic);
+
+        while (it.hasNext()) {
+            MessageAndMetadata messageAndMetadata = it.next();
+            String ip = (String) messageAndMetadata.key();
+
+            // Once we release a new version of the avro deserializer that can return SpecificData, the deep copy will be unnecessary
```
ewencp: We should figure out how we want to manage this process. Would it make sense to have another branch with these examples coded as we would with the new release, or should we just update after the release? Things I see that could be changed:

1. New consumer! Much improvement! So wow! doge
2. ConsumerConfig constants can be used instead of using strings (i.e. `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` instead of `"auto.offset.reset"`).
3. The change noted here where we'll be able to deserialize directly to the `SpecificRecord`

specific-avro-consumer/src/main/java/io/confluent/examples/consumer/AvroClicksSessionizer.java
```diff
@@ -0,0 +1,152 @@
...
+
+        // hardcoding the Kafka server URI for this example
+        props.put("bootstrap.servers", "localhost:9092");
+        props.put("acks", "all");
+        props.put("retries", 0);
+        props.put("key.serializer", "io.confluent.kafka.serializers.KafkaAvroSerializer");
+        props.put("value.serializer", "io.confluent.kafka.serializers.KafkaAvroSerializer");
+        props.put("schema.registry.url", url);
+
+        KafkaProducer<String, LogLine> producer = new KafkaProducer<String, LogLine>(props);
```
ewencp: I realized this is using whatever the default target JVM settings are for maven and that means we're not using diamond syntax here, and maybe missing some other things. We might just want to specify min 1.7 and reap some of the benefits in this code. Definitely not critical though, just a thought.

ewencp:
@gwenshap Left a couple more comments, but they are all minor or can be addressed as follow-ups. Feel free to merge!

specific-avro-consumer/README.md
gwenshap: Good catch. Can't argue with being gender neutral :)

PR #6, opened 2015-10-06
Add cli arguement for producer to be able to use it with a remote kafka node/cluster

mquirin:


miguno:
Closing this for now because the PR is pretty old and causes conflicts with latest `master`.  @mquirin : feel free to reopen and update to the latest code in `master` if need be.

PR #9, opened 2016-01-19
Add new consumer example

Ishiihara:
This is a simple example using a queue to decouple data consumption from data processing.  The implementation is straightforward:
1. We have one consumption thread plus a number of processor thread on each machine.  
2. Putting data returned by poll() to a queue. Which will be later processed by one of the processor thread. 
3. Synchronous offset commit is used. 
3. This only works if the result does not depending on the order of data. 

@hachikuji @ewencp

### DIFF
--- a/new-consumer/README.md
+++ b/new-consumer/README.md
@@ -1,0 +1,29 @@
+Quickstart
+----------
+
+Before running the examples, make sure that Zookeeper and Kafka are running. In what follows, 
+we assume that Zookeeper and Kafka are started with the default settings.
+
+    # Start Zookeeper
+    $ bin/zookeeper-server-start etc/kafka/zookeeper.properties
+
+    # Start Kafka
+    $ bin/kafka-server-start etc/kafka/server.properties
+
+   
+First create a topic called test:
+
+    # Create test topic
+    $ bin/kafka-topics --create --zookeeper localhost:2181 --replication-factor 1 \
+      --partitions 1 --topic test
+
+Then start the new-consumer example: 
+    
+    $ mvn exec:java -Dexec.mainClass="io.confluent.examples.consumer.NewConsumerExample" \
+     -Dexec.args="5"
+
+Finally start the console producer:
+
+    $ bin/kafka-console-producer.sh --broker-list localhost:9092 --topic test
+    This is a message
+    This is another message
--- a/new-consumer/pom.xml
+++ b/new-consumer/pom.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.confluent</groupId>
+    <artifactId>new-consumer-example</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+
+    <properties>
+        <kafka.version>0.9.0.0</kafka.version>
+        <kafka.scala.version>2.11</kafka.scala.version>
+        <confluent.version>2.0.0</confluent.version>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
+    </properties>
+
+    <repositories>
+        <repository>
+            <id>confluent</id>
+            <url>${confluent.maven.repo}</url>
+        </repository>
+    </repositories>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_${kafka.scala.version}</artifactId>
+            <version>${kafka.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>2.5.1</version>
+                <inherited>true</inherited>
+                <configuration>
+                    <source>1.7</source>
+                    <target>1.7</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/new-consumer/src/main/java/io/confluent/examples/consumer/ConsumeLoop.java
+++ b/new-consumer/src/main/java/io/confluent/examples/consumer/ConsumeLoop.java
@@ -1,0 +1,142 @@
+/**
+ * Copyright 2016 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.examples.consumer;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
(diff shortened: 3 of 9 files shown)


new-consumer/README.md
```diff
@@ -0,0 +1,29 @@
...
+      --partitions 1 --topic test
+
+Then start the new-consumer example: 
+    
+    $ mvn exec:java -Dexec.mainClass="io.confluent.examples.consumer.NewConsumerExample" \
+     -Dexec.args="5"
+
+Finally start the console producer:
+
+    $ bin/kafka-console-producer.sh --broker-list localhost:9092 --topic test
```
ewencp: This is inconsistent with the other example commands -- it includes the `.sh` suffix.

new-consumer/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+    <groupId>io.confluent</groupId>
+    <artifactId>new-consumer-example</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+
+    <properties>
+        <kafka.version>0.9.0.0</kafka.version>
+        <kafka.scala.version>2.11</kafka.scala.version>
+        <confluent.version>2.0.0</confluent.version>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
```
ewencp: Probably don't need this unless we use `kafka.version=0.9.0.0-cp1`. They do differ, so we might want to use the CP version of Kafka 0.9.0.0.

new-consumer/src/main/java/io/confluent/examples/consumer/ConsumeLoop.java
```diff
@@ -0,0 +1,142 @@
...
+  private final CountDownLatch shutdownLatch;
+
+  private final Queue<ConsumerRecord<Object, Object>> queue;
+  private final State<Object, Object> state;
+  private int numProcessors;
+  private Processor[] processors;
+  private Thread[] processorThreads;
+  private boolean committed = true;
+
+  public ConsumeLoop(Properties config, List<String> topics, int numProcessors) {
```
ewencp: Javadocs explaining what this does?

new-consumer/src/main/java/io/confluent/examples/consumer/ConsumeLoop.java
```diff
@@ -0,0 +1,142 @@
...
+    this.consumer = new KafkaConsumer<>(config);
+    this.topics = topics;
+    this.shutdown = new AtomicBoolean(false);
+    this.shutdownLatch = new CountDownLatch(1);
+
+    this.numProcessors = numProcessors;
+    processors = new Processor[numProcessors];
+    processorThreads = new Thread[numProcessors];
+
+    queue = new LinkedList<>();
```
ewencp: If we have a single queue and multiple processor threads, how do we get any guarantee that records with the same keys are processed together?

new-consumer/src/main/java/io/confluent/examples/consumer/ConsumeLoop.java
```diff
@@ -0,0 +1,142 @@
...
+    for (int i = 0; i < numProcessors; ++i) {
+      processorThreads[i].start();
+    }
+  }
+
+  public void run() {
+    try {
+      consumer.subscribe(topics);
+      while (!shutdown.get()) {
+        if (processFinished()) {
```
ewencp: I'm not sure the process you're using here where you grab a batch of data, then pause the consumer and try to process it all and commit is going to work well. Seems like there could be a couple of potential problems there:

1. I'd need to check exactly how the consumer handles this, but the poll that returns data will also trigger the next fetch request. I'm not sure if pausing after that will cause the returned data to be discarded -- we might end up double-fetching data.
2. You don't know how much data is going to be returned with each fetch, and the cost of processing data could vary drastically. Seems like you could end up with a really excessive number of offset commits, which are synchronous here and therefore have a fair amount of overhead.
3. Isn't it likely to result in pretty inconsistent load on each of the processor threads? They'll get stuck waiting for the drained queue to be repopulated, but since the driver thread needs to wait for all threads to finish processing messages, any variance in processing time can cause many threads to wait for a long time on one slow thread.

new-consumer/src/main/java/io/confluent/examples/consumer/ConsumeLoop.java
```diff
@@ -0,0 +1,142 @@
...
+            consumer.commitSync();
+            committed = true;
+          }
+
+          for (TopicPartition tp : consumer.assignment()) {
+            consumer.resume(tp);
+          }
+        }
+
+        ConsumerRecords<Object, Object> records = consumer.poll(2000);
```
ewencp: If all partitions are paused, what's the point of calling `poll()` here?

new-consumer/src/main/java/io/confluent/examples/consumer/ConsumeLoop.java
```diff
@@ -0,0 +1,142 @@
...
+      for (int i = 0; i < numProcessors; ++i) {
+        finished = finished && processors[i].isProcessFinished();
+      }
+      return finished;
+    }
+  }
+
+  public void shutdown() throws InterruptedException {
+    log.info("Shutting down consumer thread");
+    shutdown.set(true);
```
ewencp: Should we also `consumer.wakeup()`? No need to wait for `poll()` to timeout if its just waiting for data.

new-consumer/src/main/java/io/confluent/examples/consumer/NewConsumerExample.java
```diff
@@ -0,0 +1,81 @@
...
+ * limitations under the License.
+ */
+package io.confluent.examples.consumer;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+
+public class NewConsumerExample {
+  private static final String CLIENT_ID_CONFIG = "client.id";
```
ewencp: Can we not get these directly from `ConsumerConfig`? The config names should be public.

new-consumer/src/main/java/io/confluent/examples/consumer/Processor.java
```diff
@@ -0,0 +1,85 @@
...
+        ConsumerRecord<K, V> record;
+        synchronized (queue) {
+          while (queue.isEmpty()) {
+            try {
+              queue.wait();
+            } catch (InterruptedException e) {
+              // ignore
+            }
+          }
+          record = queue.poll();
```
ewencp: This needs to handle null return values. Since you use `notifyAll`, you might insert a single record, wake up 4 threads, and they could all see a non-empty queue. They'd exit the synchronized block and then all try to `poll()`, but only one could get a record back.

I'm wondering if this code could be significantly simplified just by using a BlockingQueue with timeouts or interrupting operations as necessary for shutdown.

new-consumer/src/main/java/io/confluent/examples/consumer/State.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.examples.consumer;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+public interface State<K, V> {
```
ewencp: Do we really need the State interface for an example like this? Why doesn't the processor just hold its state directly? Does this class add value somehow? I see it is passed into the processor constructors in ConsumeLoop, but ConsumeLoop doesn't do anything with it so it seems like the relevant state could have just been allocated in the RecordCounter.

dmatrix:
Was this inspired by the Reactive Stream’s strategy and scheme of bounded queue and back pressure?

Cheers
Jules

--
The Best Ideas Are Simple
Jules S. Damji
e-mail:dmatrix@comcast.net
e-mail:jules.damji@gmail.com
twitter:@2twitme

ewencp:
@Ishiihara This seems like a rather complex example -- one of the benefits of the new consumer is that you can write pretty simple, single-threaded consumers and parallelize pretty easily just by using multiple consumer instances.

Examples like this definitely have a place (with good explanations of when this approach is useful and the important drawbacks!), but I'd probably move the entire example to a directory with a different name, e.g. multithreaded-new-consumer or something. I think the most basic new-consumer example should probably be as simple as possible, or where it becomes more complicated, demonstrate important features (e.g. rebalance callbacks, sync/async offset commit, etc.)

miguno:
@ewencp @Ishiihara : What shall we do with this very old PR?

Ishiihara:
@miguno I am closing it for now. Will revisit later.

PR #16, opened 2016-03-17
Solving compiling problem (java compile version)

asnir:
Adding maven-compiler-plugin

PR #18, opened 2016-03-17
Solving compiling problem (java compile version)

asnir:
Adding maven-compiler-plugin


kafka-clients/specific-avro-consumer/pom.xml
```diff
@@ -38,6 +38,7 @@
         <confluent.version>1.0</confluent.version>
         <avro.version>1.7.7</avro.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <java.version>1.8</java.version>
```
miguno: In this sub-project we still support 1.7, so we can't switch to 1.8 I'm afraid.  See #16.

miguno:
Closed because we can't switch from Java 1.7. to 1.8 at the moment.

PR #21, opened 2016-04-12
MINOR: Added test for compact changelog topics

enothereska:


enothereska:
@miguno @guozhangwang could you have a look, but do "not" merge. This is the integration test we need for KAFKA-3504 https://issues.apache.org/jira/browse/KAFKA-3504 .

miguno:
@enothereska :  My first feedback would be to not "pollute" the existing integration tests -- whose purpose is to demonstrate certain use cases -- with tests that verify whether topics are actually being compacted when they should.  Instead, I'd prefer a separate, special-purpose integration test whose only purpose is to assert exactly this (compaction) behavior.

enothereska:
@miguno yup, that's the plan, hence the "do not merge" note.

kafka-streams/src/test/java/io/confluent/examples/streams/kafka/KafkaEmbedded.java
```diff
@@ -157,4 +161,65 @@ public void createTopic(String topic,
...
+
+    Map<String, Properties> topicConfigs = AdminUtils.fetchAllTopicConfigs(zkUtils);
+    Iterator it = topicConfigs.iterator();
+    while (it.hasNext()) {
+      Tuple2<String, Properties> topicConfig = (Tuple2<String, Properties>) it.next();
+      String topic = topicConfig._1;
+      Properties prop = topicConfig._2;
+
+      // statechange logs should be compacted
+      if (topic.contains("changelog")) {
```
miguno: Is this simple string-matching sufficient to correctly identify statechange logs?  (really no false positives or false negatives?)

Issue #20, opened 2016-04-01
SpecificAvroDeserializer NullPointerException


billmag:
Hello-

I'm getting the below when trying to run an example similar to the WikipediaFeedAvroLambdaExample. My build successfully runs the com.commercehub.gradle.plugin:gradle-avro-plugin to generate the org.apache.avro.specific.SpecificRecord for my .avsc file, and everything else seems pretty reasonable. I'm also running the SchemaRegistry from the 2.1.0-alpha download of the Confluent platform.

Do you expect the SpecificAvroDeserializer to work? I get a NullPointerException on "inner", and given that the empty constructor is used and there is nothing that ever seems to set inner, I'm wondering how it's supposed to work at all? I manually registered my Avro schema with the registry via curl, and that worked fine.

The behavior I see is:

At boot:
```
log4j:WARN No appenders could be found for logger (org.apache.kafka.streams.StreamsConfig).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

Then, after I publish my avro encoded message to the topic, I immediately get a NullPointerException. I don't get any stdout output from the SchemaRegistry, so I assume there's no magic causing it to be queried.
```
Exception in thread "StreamThread-1" java.lang.NullPointerException
	at io.confluent.examples.streams.utils.SpecificAvroDeserializer.deserialize(SpecificAvroDeserializer.java:51)
	at io.confluent.examples.streams.utils.SpecificAvroDeserializer.deserialize(SpecificAvroDeserializer.java:24)
	at org.apache.kafka.streams.processor.internals.SourceNode.deserializeValue(SourceNode.java:41)
	at org.apache.kafka.streams.processor.internals.RecordQueue.addRawRecords(RecordQueue.java:77)
	at org.apache.kafka.streams.processor.internals.PartitionGroup.addRawRecords(PartitionGroup.java:113)
	at org.apache.kafka.streams.processor.internals.StreamTask.addRecords(StreamTask.java:134)
	at org.apache.kafka.streams.processor.internals.StreamThread.runLoop(StreamThread.java:334)
	at org.apache.kafka.streams.processor.internals.StreamThread.run(StreamThread.java:248)

Process finished with exit code 0
```

It would be good to know if the answer is just to wait for a later release of kafka-streaming, or if I should go about trying to get this to work (or pursue a different deserialization method)?

miguno:
[...]

PR #23, opened 2016-04-15
GH-20: fix NPE in Avro serdes when running src/main/ Streams examples

miguno:
This is to address https://github.com/confluentinc/examples/issues/20.

@ewencp: Looks good to you?  /cc @guozhangwang

### DIFF
--- a/kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
+++ b/kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
@@ -17,6 +17,8 @@ package io.confluent.examples.streams;
 
 import io.confluent.examples.streams.utils.GenericAvroDeserializer;
 import io.confluent.examples.streams.utils.GenericAvroSerializer;
+import io.confluent.kafka.serializers.KafkaAvroDeserializerConfig;
+
 import org.apache.avro.generic.GenericRecord;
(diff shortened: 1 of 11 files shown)


kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
```diff
@@ -50,6 +50,8 @@ public static void main(String[] args) throws Exception {
         streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
         // Where to find the corresponding ZooKeeper ensemble.
         streamsConfiguration.put(StreamsConfig.ZOOKEEPER_CONNECT_CONFIG, "localhost:2181");
+        // Where to find the Confluent schema registry instance(s)
+        streamsConfiguration.put("schema.registry.url", "http://localhost:8081");
```
ewencp: We've been bad about this elsewhere (I think because original the constant wasn't properly made public), but we should use the constant from, e.g., `KafkaAvroDeserializerConfig` like we do with the `StreamsConfig` constants.

ewencp:
@miguno Having not actually executed anything, the changes look fine to me.

miguno:
Thanks @ewencp .  I replaced the hardcoded "schema.registry.url" with a reference to the config variable.

PR #26, opened 2016-05-10
KSTREAMS-146: Added auto reset config param 

enothereska:


enothereska:
@guozhangwang @miguno this is a WiP PR, could you check just this first example please?

miguno:
Can you confirm:

The only reason we need this change (auto.offset.reset to Earliest) is because, in the javadocs of the example, we tell people to create+populate the input topic *first*, right?  That is, if we tell people: 1. create the input topic, 2. start the Streams example, 3. write some data to the input topic -- then we don't need this change.

```java
 * Before running this example you must create the source topic (e.g. via
 * `kafka-topics --create ...`) and write some data to it (e.g. via
 * `kafka-console-producer`). Otherwise you won't see any data arriving in the output topic.
```

enothereska:
@miguno confirmed. So I suppose we can change the Javadoc, what do you suggest?

miguno:
Some food for thought however, now that I thought a little bit more about it.  The TL;DR:  Perhaps update the javadocs to be:  1. create topic, 2. start Streams app, 3. write input data to input topic.

Why?  At the moment we don't have a good "restart everything from scratch" setup (e.g. due to not having yet a convenient way to reset consumer offsets from previous runs).  My fear with your change -- now that I have thought a bit more about it -- is that users might think that, after their first run, re-running the application/demo will re-read all previous input data from the input topic.  But this won't happen.  `auto.offset.reset` only triggers if NO consumer offsets whatsoever are available for a consumer group (here: the Streams app, where the consumer group id is derived from `application.id`).  So adding this option might trick people into expecting a behavior that does not exist (yet).

WDYT?

enothereska:
Yeah, I'll change JavaDoc for now, makes sense. Thanks.

PR #35, opened 2016-06-17
Fix PageViewRegion Examples

dguy:
@miguno fixes for the PageViewRegion Examples

### DIFF
--- a/kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionExample.java
+++ b/kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionExample.java
@@ -22,6 +22,7 @@ import io.confluent.kafka.serializers.AbstractKafkaAvroSerDeConfig;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.streams.KafkaStreams;
@@ -37,6 +38,7 @@ import org.apache.kafka.streams.kstream.ValueMapper;
 import org.apache.kafka.streams.kstream.Windowed;
 
 import java.io.File;
+import java.io.InputStream;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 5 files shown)


miguno:
Can we also add instructions for writing input data with kafka avro console producer? And to read output via k-a-c-consumer?

miguno:
Also, for Avro: where/how to access the schema files when running the cli commands because we are not assuming that schema registry is running to help with that. Alternatively we can include schema registry. But then we ahould also explain how to start and stop it. 

Perhaps we can point to the CP "main" quickstart for starting/stopping Kafka and ZK, and if we also want to leverage schema registry, then we can also point to the CP regy quickstart. Though I would only add these pointers as "for further details", and perhaps include the few commands to start broker, ZK, etc (in different terminals) directly in the examples. Idea: explain this centrally in kafka-streams/README , then refer to the README in the example's top javadocs. 

WDYT?

miguno:
Ah sorry. I missed the driver which writes Avro data.

miguno:
So other than the suggestion to also include broker/ZK instructions for start and stop this PR is +1 for me.

kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionExample.java
```diff
@@ -48,9 +50,21 @@
  * "PageViews" with a user profile table that reads from a topic named "UserProfiles" to compute the
  * number of page views per user region.
  *
- * Note: Before running this example you must 1) create the source topic (e.g. via
- * `kafka-topics --create ...`), then 2) start this example and 3) write some data to
- * the source topic (e.g. via `kafka-avro-console-producer`). Otherwise you won't see any data
+ * Note: Before running this example you must
+ * 1) Start Zookeeper, Kafka and Schema Registry
+ * 2) create the topics e.g.
+ * bin/kafka-topics --create --topic PageViews --zookeeper localhost:2181 --partitions 1
```
miguno: We should add that these commands (ie without .sh extension) are for CP 3.0.0 only.  Apache Kafka scripts have the .sh extension.

kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.Properties;
+import java.util.Random;
+import java.util.stream.IntStream;
+
+import io.confluent.examples.streams.utils.WindowedStringDeserializer;
+
+public class PageViewRegionExampleDriver {
```
miguno: We should also explain how to run this driver, eg when using the packaged jar (from mvn package).

miguno: The explanation can be here or directly in the two PageView examples.

dguy:
Made some further doco changes. I'm going to merge this in as is. We can always add more later.

PR #36, opened 2016-06-18
make change to 2nd case so that it publishes to topic and add std Con…

phderome:
…sumerConfig AUTO_OFFSET_RESET_CONFIG (earliest) setting and a bit of usage comments.

### DIFF
--- a/kafka-streams/src/main/java/io/confluent/examples/streams/MapFunctionLambdaExample.java
+++ b/kafka-streams/src/main/java/io/confluent/examples/streams/MapFunctionLambdaExample.java
@@ -15,6 +15,7 @@
  */
 package io.confluent.examples.streams;
 
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.streams.KafkaStreams;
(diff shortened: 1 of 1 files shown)


phderome:
Not sure where all the errors come from as I am not that familiar as to how to integrate kafka-examples with full Kafka dependencies properly. In event, mvn -DskipTests=true package worked fine in my local copy.

miguno:
Thanks for this PR, @phderome!

For legal reasons we need you to sign the Contributor License Agreement (see above), sorry for having to ask for this.

kafka-streams/src/main/java/io/confluent/examples/streams/MapFunctionLambdaExample.java
```diff
@@ -32,6 +33,17 @@
  * Use cases include e.g. basic data sanitization, data anonymization by obfuscating sensitive data
  * fields (such as personally identifiable information aka PII).
  *
+ * ./bin/kafka-console-consumer --zookeeper localhost --topic UppercasedTextLinesTopic
```
miguno: What about also explaining how to write some example input data via `./bin/kafka-console-producer` while we're at it?

phderome:
i was trying to sign it at the time but had usage difficulty. In all
honesty, I think Damian and team will have things very well covered and may
not find this PR all that useful. I had initiated it prior to your
commitment for fixes in short term.

So, I think we can just let this PR die.

miguno:
@phderome : Ok, still leave it open for the moment.  Also:  The test failures above are not related to your code changes (i.e. no mistake on your part).

phderome:
I just signed as individual contributor.


Nice to hear about no mistakes on my part, ok.

miguno:
Closed because the MapLambdaFunctionExample has already been improved (with even more detailed step-by-step instructions).

PR #38, opened 2016-06-21
Fixes/improvements for exampes: map, pageview, anomaly detection, sum, wordcount

miguno:
Ping anyOf(@dguy @mjsax @enothereska @guozhangwang).

I'll follow-up separately for the remaining examples (top articles, user region, wikipedia feeds examples).

### DIFF
--- a/kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
+++ b/kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
@@ -1,24 +1,19 @@
 /**
  * Copyright 2016 Confluent Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
  *
  * http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
  */
 package io.confluent.examples.streams;
 
-import io.confluent.examples.streams.utils.GenericAvroSerde;
-import io.confluent.kafka.serializers.KafkaAvroDeserializerConfig;
-
-import org.apache.avro.generic.GenericRecord;
+import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.KeyValue;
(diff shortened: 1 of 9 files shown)


kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionLambdaExample.java
```diff
@@ -47,40 +47,90 @@
...
+ * $ bin/kafka-console-consumer --zookeeper localhost:2181 --topic PageViewsByRegion --from-beginning \
+ *          --property value.deserializer=org.apache.kafka.common.serialization.LongDeserializer
+ * }
+ * </pre>
+ *
+ * You should see output data similar to:
+ *
+ * <pre>
+ * {@code
+ * 1
```
miguno: I am aware that this output (which shows record values only) isn't that great.  The problem we have is that the record keys of this example are of type `Windowed<String>`, for which there's no out-of-the-box deserializer available that we can pass to `kafka-console-consumer`.

miguno: What I did in the anomaly detection example was to add some additional operations to the topology whose only purpose is to massage the output data in a way that's easier to parse/show by kafka-console-consumer.  Perhaps we also want to do this here?

dguy: Yes - i did something similar in some tests when i wanted to pipe a Windowed stream to a non-windowed stream. i.e., map the Window to a String.

dguy: Thought i'd already commented here (must have been dreaming!). Yes can probably map the keys of the Windowed stream to Strings

dguy:
LGTM

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
```diff
@@ -30,49 +25,144 @@
...
+ * #
+ * # Every line you enter will become the value of a single Kafka message.
+ * $ bin/kafka-console-producer --broker-list localhost:9092 --topic UserClicks
+ * }
+ * </pre>
+ *
+ * 5) Inspect the resulting data in the output topics, e.g. via `kafka-console-consumer`. Note that
+ * it may take a while until you see some output in the console because, by default, the example
+ * application will operate on 1-minute tumbling windows (so initially you must wait for one minute
+ * to see the first output).
```
mjsax: No intermediate output for each window update?

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
```diff
@@ -30,49 +25,144 @@
...
+ * }
+ * </pre>
+ *
+ * You should see output data similar to:
+ *
+ * <pre>
+ * {@code
+ * alice   3
+ * alice   4
+ * }
```
mjsax: Same question as above: no intermediate results?

kafka-streams/src/main/java/io/confluent/examples/streams/MapFunctionLambdaExample.java
```diff
@@ -26,13 +24,86 @@
 import java.util.Properties;
 
 /**
- * Demonstrates how to perform simple, state-less transformations via map functions.
- * See also the Scala variant MapFunctionScalaExample.
+ * Demonstrates how to perform simple, state-less transformations via map functions. See also the
+ * Scala variant MapFunctionScalaExample.
```
mjsax: `{@link MapFunctionScalaExample}`

kafka-streams/src/main/java/io/confluent/examples/streams/MapFunctionLambdaExample.java
```diff
@@ -26,13 +24,86 @@
...
+ * running example application (step 3) will automatically process this input data and write the
+ * results to the output topics.
+ *
+ * <pre>
+ * {@code
+ * # Start the console producer.  You can then enter input data by writing some line of text,
+ * # followed by ENTER:
+ * #
+ * #   hello kafka streams<ENTER>
+ * #   all streams lead to kafka<ENTER>
```
mjsax: Nice one :)

kafka-streams/src/main/java/io/confluent/examples/streams/SumLambdaExample.java
```diff
@@ -24,31 +24,87 @@
...
- * 2) create the topics e.g.
- * bin/kafka-topics --create --topic numbers-topic --zookeeper localhost:2181 --partitions 1 --replication-factor 1
- * bin/kafka-topics --create --topic sum-of-odd-numbers-topic --zookeeper localhost:2181 --partitions 1 --replication-factor 1
- * (Note the above commands are for CP3.0.0 only. For Apache Kafka it should be `bin/kafka-topics
- * .sh ...`)
- *
- * 3) start this example either in your IDE or on the command line. If via the command line please
- * refer to:
- * <a href='https://github.com/confluentinc/examples/tree/master/kafka-streams#packaging-and-running'>Packaging</a>
+ * Demonstrates how to use `reduceByKey` to sum numbers. See `SumLambdaIntegrationTest` for an
```
mjsax: `{@link SumLambdaIntegrationTest}`

mjsax:
Great work @miguno !

kafka-streams/src/main/java/io/confluent/examples/streams/SumLambdaExample.java
miguno: Can't do that unfortunately because `SumLambdaIntegrationTest` is under `src/test`, so this class (under `src/main`) isn't aware of it and thus can't link to it. :-(

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
miguno: Doh, left-over text from an earlier draft (when I was still running into issues with the pipeline).

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
miguno: Same question, but different context and thus different answer. :-)

Yes, there are intermediate results.  `alice -> 3` is an intermediate result, followed by `alice -> 4`.

And no (kinda) because the last `filter` step in the topology removes all records that have null values.  This step is required to prevent LongDeserializer (in the console) from failing, and has the side effect that you won't see any results for "bob" and "charlie", and won't see results for `alice -> 1` (the 1 is actually a `null`) and `alice -> 2` (the 2 is actually a `null`).

kafka-streams/src/main/java/io/confluent/examples/streams/MapFunctionLambdaExample.java
miguno: Fixed.

kafka-streams/src/main/java/io/confluent/examples/streams/SumLambdaExample.java
```diff
@@ -24,31 +24,87 @@
...
+ * # Here: Write input data using the example driver.  Once the driver has stopped generating data,
+ * # you can terminate it via `Ctrl-C`.
+ * $ java -cp target/streams-examples-3.0.0-standalone.jar io.confluent.examples.streams.SumLambdaExampleDriver
+ * }
+ * </pre>
+ *
+ * 5) Inspect the resulting data in the output topics, e.g. via `kafka-console-consumer`. Note that
+ * it may take a while until you see some output in the console because, by default, the example
+ * application will operate on 1-minute tumbling windows (so initially you must wait for one minute
+ * to see the first output).
```
mjsax: Same here and in some other examples, too. (sorry for not marking it)

miguno: Fixed.  I didn't spot any further places.

guozhangwang:
This is really great, thanks @miguno !

PR #52, opened 2016-08-04
update pom to use 3.0.1-SNAPSHOT and 0.10.0.1-SNAPSHOT

dguy:


ewencp:
What is the meaning of a branch with kafka 0.10.0.x naming but fixed cp-3.0.1 naming? Why are examples bumping absolute release versions but Kafka and Confluent versions switching to SNAPSHOT versions? Normally we'd match CP/Kafka versions as either SNAPSHOT or release versions. /cc @norwood

dguy:
It is currently a place holder for the next release. We can rename the branch to whatever we want. I've set the versions to SNAPSHOT for now so it can be built against 0.10.0.1 & 3.0.1, plan being to remove the SNAPSHOT once we have a release

dguy:
Doesn't build anyway, so i'll just hold off until release and update then

PR #58, opened 2016-09-01
Add `through()` step whenever we re-key data

miguno:


miguno:
Please review anyOf @dguy @enothereska @guozhangwang @hjafarpour @mjsax.

Note that two things happen because of re-partitioning, and please double-check my reasoning here.

1. Order of data: Some aggregations result in a different processing order.  For example, we have a change from `(europe, 69)` to `(europe, 53)` in the `Join*IntegrationTests` because now the two records from Alice are processed before Dave's record is processed.  That is, before (with partition=1), we had `alice -> dave -> alice`, now we have `alice -> alice -> dave`.
2. Asserts: Because of (1) we must switch from `containsExactlyElementsOf()` to `containsOnlyElementsOf()`.  Hence different test runs may result in differently ordered output data, depending on which key is being processed first.

kafka-streams/src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
```diff
@@ -52,7 +52,7 @@
 
   @BeforeClass
   public static void startKafkaCluster() throws Exception {
-    CLUSTER.createTopic(inputTopic);
+    CLUSTER.createTopic(inputTopic, 2, 1);
```
miguno: Added this to ensure that, even though because the topology uses `selectKey()`, we validate that no re-partitioning is needed.

dguy:
Your reasoning seems sound to me. LGTM

PR #59, opened 2016-09-01
add shutdown hooks to examples

dguy:


dguy:
@miguno @enothereska @mjsax

miguno:
LGTM.

I suppose we'll also update the other branches?

dguy:
yes will update other branches, too

PR #61, opened 2016-09-11
HOTFIX: getting examples ready for KIP-63

enothereska:


enothereska:
@dguy @miguno please have a look. Thanks.

dguy:
LGTM

miguno:
@enothereska : Why do we need this setting?  Also, should we actually add a code comment why we need that setting?

enothereska:
@miguno we haven't usually added code comments in the integration tests about the settings so it might look odd to single this one out. 

We need to set the commit interval to fire frequently so that we do not coalesce aggregate values in the cache. That way if the test previously got K:1, K:2, K3 before it will continue to get those 3 values (rather than just K:3, which would break the assertions). We have tests in the AK brach to cover the latter case. We might want to add further examples in confluentinc, but the purpose of this commit was simply not to break the tests.

miguno:
I hear you, but my point is that, I think, the majority of the user will have no idea why changing this setting to some other value would "suddenly" cause an integration test to fail.  For example, how would they deduce that not configuring a setting called `COMMIT_INTERVAL_MS_CONFIG` would (or would not) cause coalescing of updates?

PR #70, opened 2016-10-10
HOTFIX: Pass at adding commit interval and adjusting examples accordingly

enothereska:


enothereska:
@dguy @miguno have a look please

kafka-streams/src/main/java/io/confluent/examples/streams/WordCountLambdaExample.java
```diff
@@ -98,7 +98,11 @@
  * }
  * </pre>
  *
- * You should see output data similar to:
+ * You should see output data similar to below: (note exact sequence will depend on
```
miguno: Perhaps:

> You should see output data similar to below.  Please note that the exact output sequence will depend on how fast you type the above sentences. If you type them slowly, you are likely to get each count update, e.g., `kafka 1`, `kafka 2`, `kafka 3`. If you type them quickly, you are likely to get fewer count updates, e.g., just `kafka 3`. This is because the commit interval is set to 10 seconds. Anything typed within that interval will be compacted in memory.

(e.g. removes parentheses)

kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionExample.java
```diff
@@ -159,6 +140,8 @@ public static void main(String[] args) throws Exception {
     streamsConfiguration.put(StreamsConfig.KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
     streamsConfiguration.put(StreamsConfig.VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
     streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+    // use a commit interval of 10 seconds, i.e., records should be cached for at most this time
+    streamsConfiguration.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 10 * 1000);
```
miguno: Why 10s?  To the user this may look either like an arbitrary magic number or like a deliberate setting (the exact choice of which isn't explained though).

kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionExample.java
```diff
@@ -159,6 +140,8 @@ public static void main(String[] args) throws Exception {
     streamsConfiguration.put(StreamsConfig.KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
     streamsConfiguration.put(StreamsConfig.VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
     streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+    // use a commit interval of 10 seconds, i.e., records should be cached for at most this time
```
dguy: I think this comment is misleading. The records will likely still be cached after they have been flushed.

kafka-streams/src/main/java/io/confluent/examples/streams/PageViewRegionLambdaExample.java
```diff
@@ -157,6 +138,8 @@ public static void main(String[] args) throws Exception {
     streamsConfiguration.put(StreamsConfig.KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
     streamsConfiguration.put(StreamsConfig.VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
     streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+    // use a commit interval of 10 seconds, i.e., records should be cached for at most this time
```
dguy: as above

enothereska:
@miguno I'm trying to avoid mentioning all about KIP-63 into one comment. 10 seconds because the test's final timeout is just 30 seconds, so we want the answer before that. In some examples it depends on how fast people type. If they don't enter data fast enough in the producer they won't see any caching effects. In other cases like the music one, 500ms is so that users can quickly get the latest top-5 charts.

miguno:
> 10 seconds because the test's final timeout is just 30 seconds, so we want the answer before that.

@enothereska: Then we should e.g. write: "The commit interval must be lower than this integration test's final [?] timeout to ensure we observe processing results before this timeout."

Similar comment for the other settings.  We don't need to explain the world, but if it IS important why a config is (a) changed from the default (which, in our case, is the case for all examples in this PR) and (b) it matters if the config is 10s (and not `1min` or ...), then we should provide a brief explanation.

enothereska:
Thanks @miguno @dguy

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
```diff
@@ -125,6 +124,9 @@ public static void main(String[] args) throws Exception {
     // Specify default (de)serializers for record keys and for record values.
     streamsConfiguration.put(StreamsConfig.KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
     streamsConfiguration.put(StreamsConfig.VALUE_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
+    // Set the commit interval to 500ms so that any changes are flushed frequently. The low latency
```
miguno: Nice!

PR #80, opened 2016-11-08
WIP: Use Kafka trunk (0.10.2-SNAPSHOT)

miguno:


miguno:
@mjsax: For some reason `ApplicationResetIntegrationTest` is now stuck in the first while-loop:

```java
    //
    // Step 4: Reset application.
    //

    // wait for application to be completely shut down
    final AdminClient adminClient = AdminClient.createSimplePlaintext(CLUSTER.bootstrapServers());
    while (!adminClient.describeConsumerGroup(applicationId).consumers().isEmpty()) {
      Utils.sleep(50);
    }
```

Does that ring a bell?

mjsax:
@miguno Yes, it does ring a bell. The correct fix should be:
`adminClient.describeConsumerGroup(groupId).consumers().get().isEmpty()` (there must be an additional `get()`).

miguno:
@mjsax: In AK, `StreamsResetter` says (0.10.1 branch):

```java
            if (!adminClient.describeGroup(groupId).members().isEmpty()) {
                throw new IllegalStateException("Consumer group '" + groupId + "' is still active. " +
                    "Make sure to stop all running application instances before running the reset tool.");
            }
```

Does this need updating, too?

miguno:
Sorry, closed accidentally.

miguno:
Pushed the fix for the hanging app reset test, but the code still isn't working.  Failing with:

```
java.lang.AssertionError: Expected 3 but received only 0 records before timeout 30000 ms
```

The resetter completes successfully, but when re-running the app it doesn't reprocess.  I will not have time for the moment to investigate this further unfortunately.  Mind taking a look @mjsax ?

mjsax:
It's a bug in `StreamsResetter` introduced here: https://github.com/apache/kafka/commit/b2fa0d75df6ae2959da8c3dbe7d83f692043e577

I am just wondering, why the AK test for StreamsResetter, did not detect it in the first place... Investigating.

mjsax:
Bug fix PR: https://github.com/apache/kafka/pull/2138

miguno:
@mjsax 
> Should we add a second test to AK?

Yes, this is a good idea.

mjsax:
@miguno Test added.

miguno:
`KAFKA-3825` added overloaded methods for `KGroupedStream#reduce()`.  As a result, the original Scala example we had for joins (`JoinScalaIntegrationTest`) stopped working: the compiler complained about ambiguity for overloaded `reduce()` methods.  What stopped working was Scala-Java interop: before, we could simply pass an anonymous Scala func into `reduce()`, and via SAM it would be turned into a `Reducer[V]` instance for Streams.  The overloading of `reduce()` broke that.

I fixed that in https://github.com/confluentinc/examples/pull/80/commits/9a801baa813c79473e4d845ddccfaa5d0606f87b for `confluentinc/examples:master` by:

1. adding a new implicit conversion that takes a Scala binary function and converts it into a `Reducer[V]` and
2. adding type ascription for the output of the anon func we pass to `reduce()`.

miguno:
Ping any of @mjsax @guozhangwang @enothereska @dguy.  Compiles and tests successfully now with latest Kafka `trunk` as of today.  Note that you must have a local install of Kafka trunk for this PR (fwiw the Travis build fails because it does not have such a local install).

kafka-streams/src/test/scala/io/confluent/examples/streams/JoinScalaIntegrationTest.scala
```diff
@@ -158,7 +159,7 @@ class JoinScalaIntegrationTest extends AssertionsForJUnit {
         .map((user: String, regionWithClicks: (String, JLong)) => new KeyValue[String, JLong](regionWithClicks._1, regionWithClicks._2))
```
ijuma: Can you not use `.map { case (user, (region, clicks)) => new KeyValue(region, clicks) }`? You may need some type annotations, but the main idea is to avoid the `_.1` and `_.2`.

enothereska:
@miguno when did `JoinScalaIntegrationTest`stop working? I'm asking since we have a jenkins job for kafka-streams/examples and it didn't complain. Wondering if it's missing something.

enothereska:
Otherwise LGTM.

miguno:
@enothereska: `JoinScalaIntegrationTest` stopped working when run against Kafka `trunk`.

kafka-streams/src/test/scala/io/confluent/examples/streams/JoinScalaIntegrationTest.scala
miguno: As @ijuma and I discussed offline, unfortunately it turns this isn't easy.  The straight-forward map/case idea sounds fails for the same reason as described e.g. at https://ci.apache.org/projects/flink/flink-docs-release-1.1/apis/scala_api_extensions.html#accept-partial-functions.

It does work with explicit types along the lines of:

```scala
.map [String, (String, Long), KeyValue[String, Long]] { case (user: String, (region: String, clicks: Long)) => ... }
```
but this is quite a mouthful, too.

In short, a more idiomatic Scala wrapper might be the better way to go (eventually).  In the meantime we have to suffer from some limitations such as these when using Scala.

PR #82, opened 2016-11-09
added missind backslash in JavaDoc

mjsax:

PR #84, opened 2016-11-09
added missind backslash in JavaDoc

mjsax:


mjsax:
Follow up on #82 (merging)

PR #94, opened 2017-02-03
Add call to cleanup() to all examples

mjsax:


mjsax:
@gwenshap @miguno @enothereska @dguy @guozhangwang

mjsax:
Build passed locally.

enothereska:
Looks a bit strange to have cleanup before start, but I understand why. I'd be curious to hear @miguno 's comments.

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
```diff
@@ -151,6 +151,10 @@ public static void main(final String[] args) throws Exception {
     anomalousUsersForConsole.to(stringSerde, longSerde, "AnomalousUsers");
 
     final KafkaStreams streams = new KafkaStreams(builder, streamsConfiguration);
+    // clean local state: should not be added in production (compare ApplicationResetExample.java)
+    // required to reset application for a re-run using bin/kafka-streams-application-reset
+    // save to call even without actual reset -- only performance penalty due to necessary state recreation
```
dguy: save -> safe

kafka-streams/src/main/java/io/confluent/examples/streams/GlobalKTablesExample.java
```diff
@@ -95,6 +94,10 @@ public static void main(String[] args) {
     final KafkaStreams
         streams =
         createStreams("localhost:9092", "http://localhost:8081", "/tmp/kafka-streams-global-tables");
+    // clean local state: should not be added in production (compare ApplicationResetExample.java)
+    // required to reset application for a re-run using bin/kafka-streams-application-reset
+    // save to call even without actual reset -- only performance penalty due to necessary state recreation
```
dguy: save -> safe

elsewhere, too

kafka-streams/src/main/java/io/confluent/examples/streams/GlobalKTablesExample.java
```diff
@@ -13,6 +13,12 @@
  */
 package io.confluent.examples.streams;
 
+import io.confluent.examples.streams.avro.Customer;
```
miguno: These changed lines indicate to me that you're using a different code formatting style?  The `org.apache` ones (= "upstream" libs) should be listed before local ones = ones from this repo.

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
```diff
@@ -151,6 +151,10 @@ public static void main(final String[] args) throws Exception {
     anomalousUsersForConsole.to(stringSerde, longSerde, "AnomalousUsers");
 
     final KafkaStreams streams = new KafkaStreams(builder, streamsConfiguration);
+    // clean local state: should not be added in production (compare ApplicationResetExample.java)
```
miguno: I'd rephrase this because IMHO it's a bit confusing/misleading, perhaps to sth like:

```java
// Always (and unconditionally) clean local state prior to starting the
// processing topology.  We opt for this unconditional call here because
// this will make it easier for you to play around with the example when
// resetting the application for doing a re-run (via the Application Reset Tool, 
// http://docs.confluent.io/current/streams/developer-guide.html#application-reset-tool).
//
// In a production scenario you typically do not want to clean up always
// as we do here, but rather only when it is truly needed.
// See `ApplicationResetExample.java` for a production-like example.
```

My reasoning is that most users will pick these code examples as templates, then adapt/modify as needed.  So I'd rather err on the side of being more verbose but totally clear on why we are doing certain things, rather than leaving room for misinterpretation or confusion.

YMMV.

kafka-streams/src/main/java/io/confluent/examples/streams/GlobalKTablesExample.java
mjsax: Do we have a code style file I can import into Intellij?

kafka-streams/src/main/java/io/confluent/examples/streams/AnomalyDetectionLambdaExample.java
mjsax: Ack

kafka-streams/src/main/java/io/confluent/examples/streams/GlobalKTablesExample.java
mjsax: Never mind. I set it manually -- results in some import reordering for other files now -- was not consistent in the first place...

PR #106, opened 2017-03-27
Add ValidateStateWithInteractiveQueriesLambdaIntegrationTest

miguno:


kafka-streams/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -195,10 +198,77 @@ private static boolean continueConsuming(int messagesConsumed, int maxMessages)
...
+   * @param expected the expected contents of the store
+   * @param <K>      the store's key type
+   * @param <V>      the store's value type
+   */
+  public static <K, V> void assertThatOldestWindowContains(ReadOnlyWindowStore<K, V> store, Map<K, V> expected) {
+    long fromBeginningOfTimeMs = 0;
+    long toNowInProcessingTimeMs = System.currentTimeMillis();
+    for (K key : expected.keySet()) {
+      long windowCounter = 0;
+      try (WindowStoreIterator<V> iterator = store.fetch(key, fromBeginningOfTimeMs, toNowInProcessingTimeMs)) {
```
miguno: This method operates on the (perhaps incorrect) assumption that `ReadOnlyWindowStore#fetch()` has an ordering guarantee so that the first returned window (in the iterator) is always the oldest available window.  Is this actually true?  We don't specify a contract for `fetch()`.

kafka-streams/src/test/java/io/confluent/examples/streams/ValidateStateWithInteractiveQueriesLambdaIntegrationTest.java
```diff
@@ -0,0 +1,149 @@
...
+    IntegrationTestUtils.produceKeyValuesSynchronously(inputTopic, inputUserClicks, producerConfig);
+
+    //
+    // Step 3: Validate the application's state by interactively querying its state stores.
+    //
+    ReadOnlyKeyValueStore<String, Long> keyValueStore =
+        IntegrationTestUtils.waitUntilStoreIsQueryable("max-store", QueryableStoreTypes.keyValueStore(), streams);
+    ReadOnlyWindowStore<String, Long> windowStore =
+        IntegrationTestUtils.waitUntilStoreIsQueryable("max-window-store", QueryableStoreTypes.windowStore(), streams);
+    // Wait one more time so that the input data can be fully processed
```
miguno: Is there a better way?  Unfortunately, waiting based on `KafkaStreams#state() == KafkaStreams.State.RUNNING` doesn't work because I have found that the `KafkaStreams` instance may already be in `RUNNING` but the state store is still not yet queryable; that is, the instance is RUNNING but attempting to do `streams.store(...)` results in `InvalidStateStoreException` ("... migrated to a another instance").

kafka-streams/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
dguy: For each key, ordering will be guaranteed.

kafka-streams/src/test/java/io/confluent/examples/streams/ValidateStateWithInteractiveQueriesLambdaIntegrationTest.java
dguy: Unfortunately when you call `KafkaStreams.start()` it is immediately in the `RUNNING` state, which is a bit confusing as it isn't really ready. You could loop on:
`KafkaStreams#allMetadataForStore(...)` until it is none empty.

kafka-streams/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
miguno: Is this the official contract?  It's not explicitly specified in the javadocs of `fetch()`.  I'd suggest to update the javadocs (I can send a PR).  WDYT?

Also, do we have the same contract for `ReadOnlySessionStore#fetch()`?

dguy: Yes - we always search from the earliest segment to the latest.

dguy: javadoc update would be helpful. Thanks @miguno

miguno: Kafka docs PR at https://github.com/apache/kafka/pull/2751

PR #113, opened 2017-04-25
Upgrade Jersey version to fix Jersey warning about existing previous registration

miguno:
Warning message was similar to:

    org.glassfish.jersey.internal.Errors logErrors
    WARNING: The following warnings have been detected: WARNING: Cannot create new registration for component type class com.fasterxml.jackson.jaxrs.base.JsonParseExceptionMapper: Existing previous registration found for the type.

### DIFF
--- a/kafka-streams/pom.xml
+++ b/kafka-streams/pom.xml
@@ -45,7 +45,7 @@
         <chill.version>0.9.2</chill.version>
         <jetty.version>9.2.12.v20150709</jetty.version>
         <jackson.version>2.8.8</jackson.version>
-        <jersey.version>2.19</jersey.version>
+        <jersey.version>2.25</jersey.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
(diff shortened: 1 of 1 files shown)


miguno:
Tests pass, and from what I can tell this version bump should be safe.  But I don't remember why we picked the specific Jersey version in the first place (2.19 was released on 29-Jun-2015).

kafka-streams/pom.xml
```diff
@@ -45,7 +45,7 @@
         <chill.version>0.9.2</chill.version>
         <jetty.version>9.2.12.v20150709</jetty.version>
         <jackson.version>2.8.8</jackson.version>
-        <jersey.version>2.19</jersey.version>
+        <jersey.version>2.25</jersey.version>
```
miguno: This matches the Jersey version of https://github.com/confluentinc/rest-utils/blob/master/pom.xml#L53 (as of today).

enothereska:
LGTM

PR #118, opened 2017-05-11
Windoweddeduplicator

cvaliente:
instead of iterating the entire store regularly and expiring old records, the same can be achieved with a WindowStore that removes old segments automatically.

### DIFF
--- a/kafka-streams/src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
+++ b/kafka-streams/src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
@@ -149,7 +149,7 @@ public class EventDeduplicationLambdaIntegrationTest {
       timeIterator.close();
 
       // add new event or update existing event with new timestamp (to prevent expiry)
-      eventIdStore.put(eventId, context.timestamp(), context.timestamp());
+      eventIdStore.put(eventId, eventTime, eventTime);
 
       return isDuplicate;
     }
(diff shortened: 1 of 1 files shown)


cvaliente:
[clabot:check]

cvaliente:
@miguno Feel free to have a look at the eventdeduplicator that I finally settled with

kafka-streams/src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
```diff
@@ -216,6 +198,7 @@ public void shouldRemoveDuplicatesFromTheInput() throws Exception {
         .withKeys(Serdes.String()) // must match the return type of the Transformer's id extractor
         .withValues(Serdes.Long())
         .persistent()
+        .windowed(TimeUnit.MINUTES.toMillis(10), TimeUnit.MINUTES.toMillis(30), 3, false)
```
miguno: I suppose that we pick 10 minutes (= 2* 5mins) here because `maintainDurationMs` is set to 5 minutes?

cvaliente: Yeah, but in practice those values probably depend on how you want to deal with late arriving values. To reduce memory and disk usage one could decrease the size to throw away old windows more frequently at the cost of potentially missing out on deduplicating late records

miguno:
LGTM, merged!

PR #120, opened 2017-05-17
Initial commit kafka-connect-streams

ybyzek:
Examples that demonstrate various ways to get data into a Kafka topic, loaded into the Kafka Streams API `KStream`, and then some basic usage of the Kafka stream processing API


ybyzek:
[clabot:check]

dbtucker:
It looks like java/avro/model files are generated by some other process.   Should the PR instead include that generation logic rather than the artifact ?

ybyzek:
@dbtucker , great point.  That file is auto-generated by "mvn generate-sources".  It is part of `scripts/run-consoleProducer.sh` and `scripts/run-jdbcGenericAvro.sh` but it turns out they all need it now.  So, I'm going to remove the *.java and add this step to the README.md

ybyzek:
@dbtucker Actually "mvn clean package" includes the "generate-sources" phase (I didn't know that until I just tested it out).  So the change I made to remove src/main/java/avro/model/Location.java was still warranted, but explicitly adding "mvn generate-sources" was unnecessary.

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+      <artifactId>avro-maven-plugin</artifactId>
+      <version>1.7.7</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.2.4</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
```
mikebin: Is this dependency required?

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+      <artifactId>connect-api</artifactId>
+      <version>0.10.2.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
```
mikebin: Is this dependency required?

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+      <artifactId>kafka-streams</artifactId>
+      <version>${kafka.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro</artifactId>
+      <version>1.7.7</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.avro</groupId>
```
mikebin: Not sure this dependency is needed here - having it defined in the plugins section might be enough.

mikebin:
A couple of minor high level comments:

- Java packages names in inconsistent case - might want to follow Java coding guidelines (all lowercase for package names): https://google.github.io/styleguide/javaguide.html

- Might want to clean up imports in the source files, eliminating any unused imports. 

- If this project is imported into an environment like IntelliJ, code does not properly compile, because the package names do not always match the directory structure. For example, ``kafka-connect-streams/src/main/java/javaProducer/SpecificAvro/SpecificAvroDeserializer.java`` has a package name ``SpecificAvro``, should be ``javaproducer.specificavro`` (Or refactored up to the top level outside of ``javaproducer``). Same is true for several other classes.

kafka-connect-streams/src/main/java/javaProducer/SpecificAvro/SpecificAvroDeserializer.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
+import io.confluent.kafka.serializers.KafkaAvroDeserializer;
+
+import static io.confluent.kafka.serializers.KafkaAvroDeserializerConfig.SPECIFIC_AVRO_READER_CONFIG;
+
+public class SpecificAvroDeserializer<T extends org.apache.avro.specific.SpecificRecord> implements Deserializer<T> {
```
enothereska: Why do we need this? I thought we had the brand new streams avro serde?

kafka-connect-streams/src/main/java/jdbcJSON/model/LocationJ.java
```diff
@@ -0,0 +1,27 @@
+package jdbcJSON.model;
+
+public class LocationJ {
```
enothereska: nit: What does the "J" stand for? Perhaps a more descriptive class name?

kafka-connect-streams/src/main/java/jdbcJSON/serde/JsonDeserializer.java
```diff
@@ -0,0 +1,42 @@
+package jdbcJSON.serde;
+
+import com.google.gson.Gson;
+import org.apache.kafka.common.serialization.Deserializer;
+
+import java.util.Map;
+
+public class JsonDeserializer<T> implements Deserializer<T> {
```
enothereska: We don't have a vanilla JSON serde?

enothereska:
LGTM module of a few minor comments. In general I don't know why we need all the serialization/deserialization stuff since we already have such classes I thought. I could be wrong though.

kafka-connect-streams/src/main/java/javaProducer/SpecificAvro/SpecificAvroDeserializer.java
ybyzek: @enothereska what is it called and since which Kafka version?  I tested on 3.2.x

kafka-connect-streams/src/main/java/jdbcJSON/model/LocationJ.java
ybyzek: @enothereska I wanted to use both the JSON Location and Avro Location class in the same place, in order to reuse constructor class for "FunStreaming", so it's a bit of a hack.  J=JSON.  Is this too confusing?

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+  <packaging>jar</packaging>
+    <repositories>
+        <repository>
+            <id>confluent</id>
+            <url>http://packages.confluent.io/maven/</url>
+        </repository>
+    </repositories>
+  <properties>
+    <!-- Keep Confluent versions as properties to allow easy modification -->
+    <confluent.version>3.2.0</confluent.version>
```
mikebin: Consider updating to 3.2.1, 0.10.2.1

kafka-connect-streams/src/main/java/jdbcJSON/serde/JsonDeserializer.java
ybyzek: @enothereska I built the code from an example of @gwenshap  , but if we have a more Confluent-blessed way of doing it, I'd absolutely prefer that.  Got a pointer?

ybyzek:
@mikebin , thank you for your keen eye.  pom files are still a bit of guesswork for me.  I have made changes as follows:
1. Modified package names to be lowercase
2. Refactored SpecificAvro and GenericAvro to a utils package so it should now compile for IntelliJ (I did not verify this with IntelliJ but I understand what you were saying)
3. Removed dependencies: avro-maven-plugin, connect-api, connect-runtime
4. Updated versions in pom to 3.2.1, 0.10.2.1

kafka-connect-streams/pom.xml
ybyzek: Done

kafka-connect-streams/pom.xml
ybyzek: Removed

kafka-connect-streams/pom.xml
ybyzek: Removed and verified it still works

kafka-connect-streams/pom.xml
ybyzek: Removed and verified it still works

kafka-connect-streams/src/main/java/jdbcJSON/serde/JsonDeserializer.java
ybyzek: @ewencp , to @enothereska 's point, is the better way to do this is to use the Json serializers provided here: https://github.com/confluentinc/schema-registry/tree/3.2.x/json-serializer/src/main/java/io/confluent/kafka/serializers .  And if yes, what dependency do I need to pull into my pom?

kafka-connect-streams/src/main/java/javaProducer/SpecificAvro/SpecificAvroDeserializer.java
enothereska: @miguno is the serdes expert that did that, I'll let him chime in.

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,91 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 
+http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>io.confluent.souptonuts</groupId>
```
mikebin: Consider changing the group and project name

miguno:
Question: Would it be possible to base this PR against `master`, i.e. upcoming Kafka 0.11 and Confluent 3.3?  Or is there a particular reason (e.g. urgency) to base it against `3.2.x`?  I am asking because upcoming 3.3 ships with an official Avro serde that is compatible with Confluent schema registry;  also, we could even be able to include an official JSON serde, too.

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+Client application reads from the Kafka topic using Serdes.String() for key and a custom JSON Serde for the value.
+
+### Notes
+
+[KAFKA-4714](https://github.com/apache/kafka/pull/2458): with this enhancement, the simple message transform will be able to cast the type of the key as Long. Until this is implemented, in this example the key is written as a String.
+
+## Example 3: JDBC source connector with Avro -> GenericRecord
+
+Uses Kafka Connect JDBC source connector to produce Avro values, and null String keys, to a Kafka topic.
+Client application reads from the Kafka topic using GenericAvro Serde for the value and then the `map` function to convert the stream to Long and a custom class.
```
miguno: Typo: "GenericAvro Serde" ?

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+3|Moscow|100
+3|Moscow|200
+1|Raleigh|700
+```
+
+# Description of the Examples
+
+## Example 1: Kafka console producer -> String
+
+Uses command line `kafka-console-producer` to produce String keys and String values to a Kafka topic.
```
miguno: Nit: You may want to put references to Java types such as "String" into backticks so they render as code in the README.  For example, `String` rather than plain String.

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+Uses Java client to produce Long keys and SpecificAvro values to a Kafka topic.
+Client application reads from the Kafka topic using Serdes.Long() for key and SpecificAvro Serde for the value.
+
+## Stream processing
+
+This repo includes a custom Java class called `FunStreaming` that demonstrates the Kafka Streams API methods `count` and `reduce`. It is run in every example above.
+
+### Notes
+
+[KAFKA-5245](https://issues.apache.org/jira/browse/KAFKA-5245): one needs to provide the Serdes twice, (1) when calling `builder.Stream()` and (2) when calling `groupByKey()`
```
miguno: - Typo: `builder.Stream()` should be `builder.stream()`.  Also, we typically write this as `KStreamBuilder#stream()`.
- Similarly, instead of `groupByKey()` we typically write `KStream#groupByKey()`, for example.

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Quickstart
+
+## Start Zookeeper, Kafka and Schema Registry
```
miguno: Two meta comments:

- Since this is for beginners, we should also explain how to install Kafka/Confluent Platform in the first place.
- Alternatively, we could instead use Confluent's Docker images instead.

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Before running the examples, make sure that Zookeeper, Kafka and Schema Registry are running. In what follows, we assume that Zookeeper, Kafka and Schema Registry are started with the default settings.
+
+```
+# Start Zookeeper
+bin/zookeeper-server-start config/zookeeper.properties
```
miguno: This and the subsequent commands assume you use Confluent Platform, we should highlight this.  For example, Apache Kafka has the name `kafka-server-start.sh` (`.sh` suffix) whereas Confluent Platform uses `kafka-server-start`.

miguno: We should also highlight from where you should run the command above.  If the user is in the wrong directory, `bin/zookeeper-server-start` won't work.

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Assumptions:
+* Broker is `localhost:9092`
+* Schema Registry is `localhost:8081`
+* table.locations is copied to `/usr/local/lib/table.locations`
+* sqlite3 database is created at `/usr/local/lib/sqlite3`
+
+To override, you will need to make appropriate changes:
+* In `scripts/`: bash scripts (file path, sqlite3 database path, ZooKeeper address, Java arguments)
+* In `files/`: properties files for Kafka connect
```
miguno: Typo: "connect" -> "Connect"

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Example 4
+./scripts/run-javaproducer.sh
+```
+
+## Overrides
+
+Assumptions:
+* Broker is `localhost:9092`
+* Schema Registry is `localhost:8081`
```
miguno: Should be: "Confluent Schema Registry"

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+* table.locations is copied to `/usr/local/lib/table.locations`
+* sqlite3 database is created at `/usr/local/lib/sqlite3`
+
+To override, you will need to make appropriate changes:
+* In `scripts/`: bash scripts (file path, sqlite3 database path, ZooKeeper address, Java arguments)
+* In `files/`: properties files for Kafka connect
+* If you're executing just the Java: `-Dexec.args="/usr/local/lib/table.locations localhost:9092 http://localhost:8081"`
```
miguno: "If you're executing just the Java":  "just the Java", what do you mean?

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+## Start Zookeeper, Kafka and Schema Registry
+
+Before running the examples, make sure that Zookeeper, Kafka and Schema Registry are running. In what follows, we assume that Zookeeper, Kafka and Schema Registry are started with the default settings.
+
+```
+# Start Zookeeper
+bin/zookeeper-server-start config/zookeeper.properties
+
+# Start Kafka
+bin/kafka-server-start config/server.properties
```
miguno: Nit: We normally prefix command lines with `$`.
```
# Start Kafka
$ bin/kafka-server-start config/server.properties
```

Also, you should tell GitHub to use bash/shell formatting for the command lines so that we benefit from syntax highlighting (e.g. code comments in gray):

```bash
# Start Kafka
$ bin/kafka-server-start config/server.properties
```

You do so by making code snippets as follows:

    ```bash
    # Note the use of ```bash instead of just ``` above
    $ code here
    ```

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Start Schema Registry
```
miguno: Should be "Start Confluent Schema Registry"

kafka-connect-streams/README.md
```diff
@@ -0,0 +1,121 @@
...
+bin/schema-registry-start config/schema-registry.properties
+```
+
+## Build the apps
+
+Compile from the pom.xml file. Generate sources to create the avro.model.Location Java code
+
+```
+mvn clean package
```
miguno: We should also explain where the packaged jar file(s) can be found.

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+  <artifactId>souptonuts</artifactId>
+  <version>1.0-SNAPSHOT</version>
```
miguno: I'd suggest to use the same version string as the Confluent version we use.  For example, if `<confluent.version>` is `3.2.1`, this version should also be `3.2.1`.  If we're talking about the `master` branch, then the various versions are typically snapshots.

Btw, we don't need to worry about not using `-SNAPSHOT` here because the code in this repository is not published to a Maven repository.

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+  <packaging>jar</packaging>
+    <repositories>
+        <repository>
+            <id>confluent</id>
+            <url>http://packages.confluent.io/maven/</url>
+        </repository>
+    </repositories>
+  <properties>
+    <!-- Keep Confluent versions as properties to allow easy modification -->
```
miguno: Nit: No need to explain this, it's obvious. :)

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    <confluent.version>3.2.1</confluent.version>
+    <kafka.version>0.10.2.1-cp1</kafka.version>
+    <avro.version>1.7.7</avro.version>
+    <!-- Maven properties for compilation -->
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+  </properties>
+  <dependencies>  
+    <!-- Add the Kafka dependency -->
```
miguno: Nit: remove this, it's obvious

kafka-connect-streams/scripts/helper-show.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+TOPIC=$1
+BOOTSTRAPSERVERS=localhost:9092
```
miguno: Nit: In bash, you'd write this as `BOOTSTRAP_SERVERS` (i.e., use underscores to separate words).

kafka-connect-streams/scripts/helper-show.sh
```diff
@@ -0,0 +1,23 @@
...
+ZOOKEEPER=localhost:2181
+
+# Run the Consumer to print the key as well as the value from the Topic
+echo -e "======= Running kafka-console-consumer for topic $TOPIC =======\n"
+timeout 5s kafka-console-consumer \
```
miguno: The command `timeout` is not available on a Mac AFAICT.  I think we should really make these scripts work on Macs, too.

kafka-connect-streams/scripts/helper-show.sh
```diff
@@ -0,0 +1,23 @@
...
+--bootstrap-server $BOOTSTRAPSERVERS \
+--from-beginning \
+--topic $TOPIC \
+--property print.key=true
+
+# View the contents of what was written to the log
+echo -e "======= Viewing log contents for topic $TOPIC =======\n"
+kafka-run-class kafka.tools.DumpLogSegments \
```
miguno: Note: This script will only work if `kafka-run-class`, `kafka-console-consumer` etc. are in the user's `PATH`.  AFAICT, there's nothing in the README/docs/... that would ensure this.

kafka-connect-streams/scripts/helper-show.sh
```diff
@@ -0,0 +1,23 @@
...
+--print-data-log \
+--files /var/lib/kafka/$TOPIC-0/00000000000000000000.log
```
miguno: Isn't this setup brittle in the way it dumps the log segments of this specific Kafka log file?  What if the log file is being rolled?  Is it important to print such lower level details?  Why not skip this step and/or use something that isn't on the level of a soldering iron and an Allen key? :)

kafka-connect-streams/scripts/run-consoleproducer.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+TABLE="locations"
+PACKAGE="consoleproducer"
+TOPIC="$PACKAGE-$TABLE"
+
+BOOTSTRAPSERVERS=localhost:9092
+
+# View contents of file
+SOURCE_FILE="files/table.$TABLE"
```
miguno: Careful:  If the user doesn't run this script as `scripts/run-consoleproducer.sh` but e.g. as `run-consoleproducer.sh` (e.g. after the user has changed into the `scripts/` directory), then `SOURCE_FILE` will not be found.

kafka-connect-streams/scripts/run-consoleproducer.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
miguno: Nit: We should rename `run-consoleproducer.sh` to `run-console-producer.sh`

kafka-connect-streams/files/jdbcgenericavro/source-quickstart-sqlite.properties
```diff
@@ -0,0 +1,31 @@
+##
+# Copyright 2015 Confluent Inc.
```
miguno: Nit: 2017.  Same for other files.

kafka-connect-streams/files/jdbcjson/connect-standalone.properties
```diff
@@ -0,0 +1,37 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
miguno: Copyright missing here.

kafka-connect-streams/README.md
miguno: Perhaps:

```
    <groupId>io.confluent</groupId>
    <artifactId>connect-streams-examples</artifactId>
```

The Kafka Streams examples in this repo use:

```
    <groupId>io.confluent</groupId>
    <artifactId>streams-examples</artifactId>
```

kafka-connect-streams/scripts/helper-connect.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "select * from $TABLE;" | sqlite3 $DB
+
+sed -i "/^connection.url/c\connection.url=jdbc:sqlite:$DB" files/$PACKAGE/source-quickstart-sqlite.properties
+
+sudo cp files/$PACKAGE/connect-standalone.properties /etc/kafka/connect-standalone.properties
+sudo cp files/$PACKAGE/source-quickstart-sqlite.properties /etc/kafka-connect-jdbc/source-quickstart-sqlite.properties
+
+sudo jps | grep ConnectStandalone | awk '{print $1;}' | xargs sudo kill -9
+
+echo "======= Running standalone Connect and sleeping 30 seconds before continuing ======="
```
miguno: Nit: says we sleep 30 seconds but we only sleep 25 :)   (if you really want to print the exact sleep time, we should use a variable for the sleep seconds)

kafka-connect-streams/scripts/helper-connect.sh
```diff
@@ -0,0 +1,26 @@
...
+sudo nohup connect-standalone /etc/kafka/connect-standalone.properties /etc/kafka-connect-jdbc/source-quickstart-sqlite.properties </dev/null &>/dev/null &
```
miguno: Why do we need sudo?

kafka-connect-streams/scripts/helper-connect.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "pragma table_info($TABLE);" | sqlite3 $DB
+echo -e "======= Viewing rows from table $TABLE in the database $DB =======\n"
+echo "select * from $TABLE;" | sqlite3 $DB
+
+sed -i "/^connection.url/c\connection.url=jdbc:sqlite:$DB" files/$PACKAGE/source-quickstart-sqlite.properties
+
+sudo cp files/$PACKAGE/connect-standalone.properties /etc/kafka/connect-standalone.properties
+sudo cp files/$PACKAGE/source-quickstart-sqlite.properties /etc/kafka-connect-jdbc/source-quickstart-sqlite.properties
+
+sudo jps | grep ConnectStandalone | awk '{print $1;}' | xargs sudo kill -9
```
miguno: Why do we need sudo?

kafka-connect-streams/scripts/helper-connect.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "======= Running standalone Connect and sleeping 30 seconds before continuing ======="
+sudo nohup connect-standalone /etc/kafka/connect-standalone.properties /etc/kafka-connect-jdbc/source-quickstart-sqlite.properties </dev/null &>/dev/null &
+sleep 25
+sudo jps | grep ConnectStandalone | awk '{print $1;}' | xargs sudo kill -9
```
miguno: Why sudo?

kafka-connect-streams/scripts/helper-show.sh
miguno: As mentioned above, this won't work on Mac OS (no `timeout` command).

kafka-connect-streams/scripts/helper-show.sh
```diff
@@ -0,0 +1,23 @@
...
+# Run the Consumer to print the key as well as the value from the Topic
+echo -e "======= Running kafka-console-consumer for topic $TOPIC =======\n"
+timeout 5s kafka-console-consumer \
+--bootstrap-server $BOOTSTRAPSERVERS \
+--from-beginning \
+--topic $TOPIC \
+--property print.key=true
+
+# View the contents of what was written to the log
+echo -e "======= Viewing log contents for topic $TOPIC =======\n"
```
miguno: Again, do we need this low-level command?

kafka-connect-streams/scripts/run-consoleproducer.sh
miguno: Nit: `run-consoleproducer.sh` -> `run-console-producer.sh`

miguno: Also, consider changing the name of this script.  It does more than running the console producer: it prints the contents of the source file, runs the console producer, compiles the Java code, runs the Java `StreamsIngest` class, and finally it runs the helper show script.

Perhaps sth like `write-data.sh`?  I'm sure there is a better name, but I haven't checked yet what `StreamsIngest` actually does.

kafka-connect-streams/scripts/run-consoleproducer.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Write the contents of the file SOURCE_FILE to the Topic `consoleproducer`, where the id is the message key and the name and sale are the message value.
+cat $SOURCE_FILE | \
+kafka-console-producer \
+--broker-list $BOOTSTRAPSERVERS \
+--topic $TOPIC \
+--property parse.key=true \
+--property key.separator='|'
+
+mvn generate-sources
```
miguno: You don't need this, it's done as part of `mvn compile`.

kafka-connect-streams/scripts/helper-connect.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "DROP TABLE IF EXISTS $TABLE;" | sqlite3 $DB
+echo "CREATE TABLE $TABLE(id INTEGER KEY NOT NULL, name VARCHAR(255), sale INTEGER);" | sqlite3 $DB
+echo ".import files/table.$TABLE $TABLE" | sqlite3 $DB
+
+echo -e "======= Viewing schema of table $TABLE in the database $DB =======\n"
+echo "pragma table_info($TABLE);" | sqlite3 $DB
+echo -e "======= Viewing rows from table $TABLE in the database $DB =======\n"
+echo "select * from $TABLE;" | sqlite3 $DB
+
+sed -i "/^connection.url/c\connection.url=jdbc:sqlite:$DB" files/$PACKAGE/source-quickstart-sqlite.properties
```
miguno: Careful: This script will not work if it isn't run as `scripts/helper-connect.sh` but e.g. as `top-level-dir/scripts/helper-connect.sh` or `helper-connect.sh`, because e.g. the relative path to `files/$PACKAGE/source-quickstart-sqlite.properties` will then be invalid.

kafka-connect-streams/scripts/helper-connect.sh
```diff
@@ -0,0 +1,26 @@
...
+
+sudo cp files/$PACKAGE/connect-standalone.properties /etc/kafka/connect-standalone.properties
```
miguno: 1. Can't we do this in a way that doesn't require `sudo`?
2. This assumes that the directory `/etc/kafka/` exists.  This is only the case IIRC if you install Kafka as RPM and/or DEB package, which means (1) it does not work for users that installed Confluent Platform via ZIP or TAR and (2) it does not work for Mac users in general.

kafka-connect-streams/scripts/helper-connect.sh
miguno: Also, since we start Connect in standalone mode and explicitly pass in the location of the properties file, the properties file doesn't need to be under `/etc/kafka-connect-jdbc/` (creating/writing to which the user will need root permissions).

kafka-connect-streams/scripts/helper-show.sh
miguno: Nit: should be `BOOTSTRAP_SERVERS`

kafka-connect-streams/scripts/run-javaproducer.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+TABLE="locations"
+PACKAGE="javaproducer"
+TOPIC="$PACKAGE-locations"
+
+BOOTSTRAPSERVERS=localhost:9092
```
miguno: Bash nit: we should separate words in variable names with underscores

kafka-connect-streams/scripts/run-javaproducer.sh
```diff
@@ -0,0 +1,23 @@
...
+SCHEMAREGISTRYURL=http://localhost:8081
+TABLELOCATIONS=/usr/local/lib/table.locations
+
+
+cp files/table.$TABLE $TABLELOCATIONS
+
+mvn generate-sources
```
miguno: Not needed, will be run as part of `mvn compile`.

kafka-connect-streams/scripts/run-javaproducer.sh
```diff
@@ -0,0 +1,23 @@
...
+mvn compile
+timeout 60s mvn exec:java -Dexec.mainClass=$PACKAGE.Driver -Dexec.args="$BOOTSTRAPSERVERS $SCHEMAREGISTRYURL $TABLELOCATIONS"
+timeout 60s mvn exec:java -Dexec.mainClass=$PACKAGE.StreamsIngest -Dexec.args="$BOOTSTRAPSERVERS $SCHEMAREGISTRYURL"
+
+scripts/helper-show.sh $TOPIC
+
+# View Schema Registry
+echo -e "======= Viewing schemas in Schema Registry for $TOPIC-value =======\n"
```
miguno: Description doesn't match the actual command, which will only print the first version of the schema.

kafka-connect-streams/src/main/java/consoleproducer/StreamsIngest.java
```diff
@@ -0,0 +1,59 @@
+package consoleproducer;
```
miguno: I'd consider changing the package names of the various java files.

One suggestion would be:

```java
package io.confluent.examples.connectandstreams
```
(Alas, Java doesn't allow for `-` in package names, so `connect-and-streams` is not valid.)

kafka-connect-streams/src/main/java/consoleproducer/StreamsIngest.java
```diff
@@ -0,0 +1,59 @@
...
+        // without well-structured data
+        KStream<Long, Location> locations = inputData.map((k, v) -> new KeyValue<Long, Location>(Long.parseLong(k), (Location) new Location (Long.parseLong(k), v.split("\\|")[0], Long.parseLong(v.split("\\|")[1])) ));
+        locations.print();
+
+        FunStreaming fun = new FunStreaming(locations);
+        fun.countMyKeys();
+        fun.aggMyValues();
+
+        KafkaStreams streams = new KafkaStreams(builder, streamsConfiguration);
+        streams.start();
```
miguno: We should add a shutdown hook.  See http://docs.confluent.io/current/streams/developer-guide.html#using-kafka-streams-within-your-application-code.

kafka-connect-streams/src/main/java/jdbcjson/model/LocationJ.java
```diff
@@ -0,0 +1,27 @@
+package jdbcjson.model;
+
+public class LocationJ {
```
miguno: Consider renaming to `LocationJSON` or sth like that.  In Java, the `J` suffix typically refers to "Java", like [AspectJ](https://eclipse.org/aspectj/).

kafka-connect-streams/src/main/java/jdbcjson/serde/WrapperSerde.java
```diff
@@ -0,0 +1,41 @@
+package jdbcjson.serde;
+
+import org.apache.kafka.common.serialization.Deserializer;
+import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.common.serialization.Serializer;
+
+import java.util.Map;
+
+
+public class WrapperSerde<T> implements Serde<T> {
```
miguno: Why do we need this?  The Streams API has built-in methods to construct a serde from a pair of serializer and deserializer.

kafka-connect-streams/src/main/java/testapi/FunStreaming.java
```diff
@@ -0,0 +1,57 @@
...
+import utils.SpecificAvroSerde;
+import avro.model.Location;
+import io.confluent.kafka.serializers.AbstractKafkaAvroSerDeConfig;
+import io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient;
+
+import java.util.Collections;
+import java.util.Map;
+
+
+public class FunStreaming {
```
miguno: Do we really need this class?  For a tutorial, I think it would be easier to include the KStream count/aggregation code directly into the various "upstream" Java classes.  No strong opinion here, but I think it's easier for the reader to follow.

kafka-connect-streams/src/main/java/utils/GenericAvroDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+package utils;
+
+import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
+import io.confluent.kafka.serializers.KafkaAvroDeserializer;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.kafka.common.serialization.Deserializer;
+
+import java.util.Map;
+
+public class GenericAvroDeserializer implements Deserializer<GenericRecord> {
```
miguno: If we can base this PR against `master`, we can leverage our new official Avro serde, and no longer need this class plus `GenericAvroSerializer` etc.

kafka-connect-streams/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,8 @@
+# Root logger option
+log4j.rootLogger=INFO, stdout
```
miguno: Might consider bumping this from `INFO` to `WARN` if the output is too chatty.  Not sure whether you intentionally want to showcase information on `INFO` level, of whether this file was simply copy-pasted.

miguno:
Many thanks for this big PR @ybyzek ! :-)

I left comments, most of which are around smoothening the UX for this new Connect-Streams example.  Notably, I think we should double-check which installation environments we'd like to support.  For example, I'd definitely consider making this work on Mac OS, not just on Linux.

kafka-connect-streams/pom.xml
ybyzek: I presume the opposite, that nothing is obvious. Not all our users are familiar with pom files; this might be their first encounter with one.

kafka-connect-streams/pom.xml
ybyzek: I presume the opposite, that nothing is obvious. Not all our users are familiar with pom files; this might be their first encounter with one.

kafka-connect-streams/scripts/helper-show.sh
ybyzek: Absolutely, thanks! I made your suggested change to follow bash naming conventions.

kafka-connect-streams/scripts/helper-show.sh
ybyzek: I added a prerequisite into the README.md

kafka-connect-streams/scripts/run-consoleproducer.sh
ybyzek: I'm going to leave it as-is because I want it to match the Java package name exactly.

kafka-connect-streams/files/jdbcgenericavro/source-quickstart-sqlite.properties
ybyzek: I'm going to leave it as-is because I want it to match the Java package name exactly.

kafka-connect-streams/scripts/helper-show.sh
ybyzek: I added a prerequisite into the README.md, so if users follow it, it will work on Mac OS

kafka-connect-streams/scripts/helper-connect.sh
ybyzek: These helper scripts aren't meant to be executed separately.  They are common code called from within the bash scripts run-*

kafka-connect-streams/scripts/helper-show.sh
ybyzek: Fixed

kafka-connect-streams/scripts/run-javaproducer.sh
ybyzek: I updated the description.

kafka-connect-streams/scripts/helper-connect.sh
miguno: Then: the `run-*` scripts suffer from the same problem.  If you call them straight from the `scripts/` directory (as `./run-consoleproducer.sh`, for example), then they will not work.

kafka-connect-streams/src/main/java/jdbcjson/serde/WrapperSerde.java
ybyzek: I got stuff “to work” but definitely need specific pointers if we want to demonstrate another “preferred” way.  What's the built-in method "to construct a serde from a pair of serializer and deserializer"?   Example please?

kafka-connect-streams/files/jdbcjson/connect-standalone.properties
ybyzek: @miguno : my file appears to match what’s in [github](https://github.com/confluentinc/kafka/blob/trunk/config/connect-standalone.properties).  If I'm looking in the wrong place, can you please provide the right link I should be looking at?

kafka-connect-streams/scripts/helper-connect.sh
ybyzek: We don't ;) Bad habit on my part to make everything sudo

kafka-connect-streams/src/main/java/jdbcjson/serde/WrapperSerde.java
miguno: It's `Serdes.from(...)`: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/serialization/Serdes.java#L147-L162

kafka-connect-streams/files/jdbcjson/connect-standalone.properties
miguno: Ah, I see.  I added my comment above because other files in this PR (I didn't check all of them) include not only the Apache license header but also:

```
# Copyright 2017 Confluent Inc.
#
# Licensed under the Apache License, Version 2.0 (the "License");
...
```

kafka-connect-streams/scripts/run-consoleproducer.sh
ybyzek: There are four examples of dataflow.  Each of the four examples has a script/run-* and a corresponding `StreamsIngest` class in each java package.  I augmented the README.md to reflect this...is it better?

kafka-connect-streams/files/jdbcjson/connect-standalone.properties
ybyzek: I added a stock comment at the top of the Java files, and a Copyright at the top of the scripts/run-* bash scripts.

kafka-connect-streams/src/main/java/jdbcjson/serde/WrapperSerde.java
ybyzek: @miguno Fantastic, thanks!!  I am using `Serdes.serdeFrom` now.  One question back to you is I have included `serde/JsonDeserializer.java` and `serde/JsonSerializer.java` in this package.  Am I required to do that or is there one included with Confluent Platform?  (or is that coming in 3.3 as well?)

kafka-connect-streams/src/main/java/utils/GenericAvroDeserializer.java
ybyzek: @miguno : here is my thought, tell me what you think.  I would like to push this out in the next 1-2 days so that people can start using it (e.g. Kai and Hans were asking).  Then when 3.3 comes out, I'll update the code with another example using the new official Avro serde and publish a blog post to go with it.

kafka-connect-streams/scripts/helper-connect.sh
ybyzek: Correct, and the README.md shows the proper way to run them.

kafka-connect-streams/scripts/helper-connect.sh
ybyzek: I know it's not ideal, but I added a note in the README.md ... in the prerequisites section.  I just can't handle all possible means of installation, all possible user file paths, etc.

kafka-connect-streams/src/main/java/consoleproducer/StreamsIngest.java
ybyzek: @miguno , I refactored it to implement your suggestion.  Thanks!!

ybyzek:
@miguno : since I changed filenames to match `io.confluent.examples.connectandstreams` package naming convention, I think some comments got "lost" from this conversation.  But basically I think the main outstanding point is whether to push to master or 3.2.x. IMHO: I'm ok with whatever enables this to push out in the next day or two, and then when 3.3 is released I can update it to use the included SpecificAvro

miguno:
As discussed offline with @ybyzek, let's keep this PR against `3.2.x` for the moment.  We can help with upgrading to 3.3/0.11 once that is out.

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/consoleproducer/StreamsIngest.java
```diff
@@ -0,0 +1,97 @@
...
+            .reduce(
+                (aggValue, newValue) -> aggValue + newValue, SALES_STORE)
+            .toStream();
+        salesAgg.print();
+
+        KafkaStreams streams = new KafkaStreams(builder, streamsConfiguration);
+        streams.start();
+
+        // Add shutdown hook to respond to SIGTERM and gracefully close Kafka Streams
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
mikebin: This code could be simplified to: `Runtime.getRuntime().addShutdownHook(new Thread(streams::close));`

ybyzek: @mikebin , thank you.  I have made the appropriate changes.

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/consoleproducer/StreamsIngest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.kafka.streams.kstream.KStream;
+import org.apache.kafka.streams.kstream.KStreamBuilder;
+import org.apache.kafka.streams.KeyValue;
+
+import io.confluent.examples.connectandstreams.avro.Location;
+
+public class StreamsIngest {
+
+    static final String INPUT_TOPIC = "consoleproducer-locations";
+    static final String INPUT_STORE = "consoleproducer-locations-store";
```
mikebin: `INPUT_STORE` is not used in this app - can be removed.

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
```diff
@@ -0,0 +1,113 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.examples.connectandstreams.javaproducer;
+
+import java.util.Properties;
+
+import org.apache.kafka.common.serialization.Serde;
```
mikebin: A lot of these imports are unused and can be removed. It looks like only the following imports are required:

```
import io.confluent.examples.connectandstreams.avro.Location;
import io.confluent.examples.connectandstreams.utils.SpecificAvroSerializer;
import io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient;
import io.confluent.kafka.serializers.AbstractKafkaAvroSerDeConfig;
import org.apache.kafka.clients.producer.KafkaProducer;
import org.apache.kafka.clients.producer.ProducerConfig;
import org.apache.kafka.clients.producer.ProducerRecord;
import org.apache.kafka.common.serialization.LongSerializer;

import java.io.BufferedReader;
import java.io.FileReader;
import java.util.*;
```

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
```diff
@@ -0,0 +1,113 @@
...
+    public static void main(String[] args) throws Exception {
+
+        if (args.length > 3) {
+            throw new IllegalArgumentException("usage: ... " +
+                "[<bootstrap.servers> (optional, default: " + DEFAULT_BOOTSTRAP_SERVERS + ")] " +
+                "[<schema.registry.url> (optional, default: " + DEFAULT_SCHEMA_REGISTRY_URL + ")] " +
+                "[<table.locations> (optional, default: " + DEFAULT_TABLE_LOCATIONS + ")] ");
+            }
+
+        final String BOOTSTRAP_SERVERS = args.length > 0 ? args[0] : DEFAULT_BOOTSTRAP_SERVERS;
```
mikebin: Non-constants like this generally are not all-uppercased. More common would be camel-cased, such as `bootstrapServers`. https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/StreamsIngest.java
```diff
@@ -0,0 +1,102 @@
...
+
+package io.confluent.examples.connectandstreams.javaproducer;
+
+import java.util.Properties;
+
+import org.apache.kafka.common.serialization.Serdes;
+
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+
+import org.apache.kafka.streams.KafkaStreams;
```
mikebin: A few imports are unneeded, complete list of needed imports below:

```
import io.confluent.examples.connectandstreams.avro.Location;
import io.confluent.examples.connectandstreams.utils.SpecificAvroSerde;
import io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient;
import org.apache.kafka.clients.consumer.ConsumerConfig;
import org.apache.kafka.common.serialization.Serdes;
import org.apache.kafka.streams.KafkaStreams;
import org.apache.kafka.streams.KeyValue;
import org.apache.kafka.streams.StreamsConfig;
import org.apache.kafka.streams.kstream.KStream;
import org.apache.kafka.streams.kstream.KStreamBuilder;

import java.util.Collections;
import java.util.Map;
import java.util.Properties;
```

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
```diff
@@ -0,0 +1,113 @@
...
+        final List<Location> locationsList = new ArrayList<Location>();
+        try (BufferedReader br = new BufferedReader(new FileReader(TABLE_LOCATIONS))) {
+            String line;
+            while ((line = br.readLine()) != null) {
+                String[] parts = line.split("\\|");
+                locationsList.add(new Location(Long.parseLong(parts[0]), parts[1], Long.parseLong(parts[2])) );
+            }
+        }
+
+        final CachedSchemaRegistryClient schemaRegistry = new CachedSchemaRegistryClient(SCHEMA_REGISTRY_URL, 100);
```
mikebin: Not sure if all this SR-specific code is required. The following code seems to work, and is simpler:

```
...
        final List<Location> locationsList = new ArrayList<>();
        try (BufferedReader br = new BufferedReader(new FileReader(TABLE_LOCATIONS))) {
            String line;
            while ((line = br.readLine()) != null) {
                String[] parts = line.split("\\|");
                locationsList.add(new Location(Long.parseLong(parts[0]), parts[1], Long.parseLong(parts[2])) );
            }
        }

        final Properties props = new Properties();
        props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, BOOTSTRAP_SERVERS);
        props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, LongSerializer.class);
        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, io.confluent.kafka.serializers.KafkaAvroSerializer.class);
        props.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, SCHEMA_REGISTRY_URL);
        KafkaProducer<Long, Location> locationProducer = new KafkaProducer<Long, Location>(props);

        locationsList.forEach(t -> {
           System.out.println("Writing location information for '" + t.getId() + "' to input topic " + INPUT_TOPIC);
           ProducerRecord<Long, Location> record = new ProducerRecord<Long, Location>(INPUT_TOPIC, t.getId(), t);
           locationProducer.send(record);
        });
...
```

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
ybyzek: Fixed

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/consoleproducer/StreamsIngest.java
ybyzek: Removed

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
ybyzek: Pruned.

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
ybyzek: Yes, I do like your code better and updated it.  However, note that my original version was from @miguno 's committed example:  https://github.com/confluentinc/examples/blob/7d14af01500d472732cfab3ad50b3200d5607b4e/kafka-streams/src/main/java/io/confluent/examples/streams/interactivequeries/kafkamusic/KafkaMusicExampleDriver.java

I don't think we lose anything, so I committed the changes....  I can roll back if anyone says otherwise.

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/StreamsIngest.java
```diff
@@ -0,0 +1,101 @@
...
+        System.out.println("Connecting to Kafka cluster via bootstrap servers " + bootstrapServers);
+        System.out.println("Connecting to Confluent schema registry at " + SCHEMA_REGISTRY_URL);
+
+        final KStreamBuilder builder = new KStreamBuilder();
+
+        Properties streamsConfiguration = new Properties();
+        streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "javaproducer-1");
+        streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+        streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+
```
mikebin: Similar to what was mentioned in `Driver.java`. This code could be simplified a bit as follows:

```
...
        streamsConfiguration.put(StreamsConfig.KEY_SERDE_CLASS_CONFIG, Serdes.LongSerde.class);
        streamsConfiguration.put(StreamsConfig.VALUE_SERDE_CLASS_CONFIG, SpecificAvroSerde.class);
        streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, SCHEMA_REGISTRY_URL);

        KStream<Long, Location> locations = builder.stream(INPUT_TOPIC);
        locations.print();

        KStream<Long,Long> sales = locations.map((k, v) -> new KeyValue<Long, Long>(k, (Long) v.get("sale")));
...
```

ybyzek: I like your simplification, but then why didn’t @miguno  write the music demo that way?  I will wait for @miguno to weigh in before making changes.

kafka-connect-streams/scripts/helper-connect.sh
mikebin: Could we just run `connect-standalone` pointing to the files in the relative path `files/$PACKAGE/...` instead of trying to copy them to `/etc/kafka/`?

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
miguno: @ybyzek just seeing this (am a bit swamped here).  Perhaps you can ping @dguy, he wrote the Kafka Music app code.

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/StreamsIngest.java
miguno: See my comment above.

kafka-connect-streams/files/jdbcgenericavro/source-quickstart-sqlite.properties
miguno: Sorry, what?  I think your response targeted a different comment?

kafka-connect-streams/pom.xml
```diff
@@ -0,0 +1,91 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 
+http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>io.confluent</groupId>
+  <artifactId>connect-streams-examples</artifactId>
+  <version>1.0-SNAPSHOT</version>
```
miguno: I think I already commented that we should consider using a version identifier that matches e.g. what we are doing in the code under `kafka-streams/` (e.g. we'd use `<version>3.2.1</version>` here).

kafka-connect-streams/scripts/helper-connect.sh
miguno: ^ What @mikebin said.  Alternatively, we could put the above under e.g. `/tmp/kafka-connect-streams/etc/kafka`.

miguno: Also, I don't like the use of `/etc/kafka` that much because it's very intrusive to the user's environment.  Just saying.  I do see the point of keeping things simple though.

Lastly, where do we even ensure that the parent dir tree `/etc/kafka` actually exists?  If it doesn't, the `cp` will fail.

kafka-connect-streams/scripts/helper-show.sh
miguno: Hmm, I don't see the changes -- have you actually pushed the commit/changes?

kafka-connect-streams/scripts/run-consoleproducer.sh
miguno: I don't see changes to the README that you talked about.  Perhaps you haven't pushed the changes yet?

kafka-connect-streams/scripts/helper-show.sh
ybyzek: @miguno you have a good eye.  I had changed them in all the OTHER files except this one ;)  Now it is fixed

kafka-connect-streams/pom.xml
ybyzek: Yep!  Updated now

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/Driver.java
miguno: I think the main difference is that @mikebin's version uses the "original" Kafka Avro serializer that we built for use with the Kafka producer client, whereas @ybyzek's version (and the one we use in the Kafka Streams examples) use the "wrapper" `SpecificAvroSerializer` we used as a means to implement a `Serde<>` for specific Avro for plugging it into Kafka Streams apps (serde = pair of serializer and deserializer).

Either approach should work here.

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/consoleproducer/StreamsIngest.java
```diff
@@ -0,0 +1,91 @@
...
+        streamsConfiguration.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 1000);
+        streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+
+        final KStream<String, String> inputData = builder.stream(Serdes.String(), Serdes.String(), INPUT_TOPIC);
+        inputData.print();
+
+        // The value is a String with another '|' separator so we need to split it again
+        // This is obviously messy but is shown just to demonstrate how messy it can be
+        // without well-structured data
+        KStream<Long, Location> locations = inputData.map((k, v) -> new KeyValue<Long, Location>(Long.parseLong(k), (Location) new Location (Long.parseLong(k), v.split("\\|")[0], Long.parseLong(v.split("\\|")[1])) ));
```
dguy: nit: you don't need the cast `(Location)`

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/consoleproducer/StreamsIngest.java
```diff
@@ -0,0 +1,91 @@
...
+        locations.print();
+
+        KStream<Long,Long> sales = locations.map((k, v) -> new KeyValue<Long, Long>(k, (Long) v.get("sale")));
```
dguy: i think you can replace `(Long) v.get("sale")` with `v.getSale()`

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/javaproducer/StreamsIngest.java
```diff
@@ -0,0 +1,102 @@
...
+        Map<String, Object> config = new HashMap<>();
+        config.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, SCHEMA_REGISTRY_URL);
+        SpecificAvroSerde<Location> locationSerde = new SpecificAvroSerde<>();
+        locationSerde.configure(config, false);
+
+        KStream<Long, Location> locations = builder.stream(Serdes.Long(), locationSerde, INPUT_TOPIC);
+        locations.print();
+
+        KStream<Long,Long> sales = locations.map((k, v) -> new KeyValue<Long, Long>(k, (Long) v.get("sale")));
```
dguy: same as comment above

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/jdbcgenericavro/StreamsIngest.java
```diff
@@ -0,0 +1,97 @@
...
+        streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "jdbcgenericavro-2");
+        streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+        streamsConfiguration.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 1000);
+        streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        streamsConfiguration.put(StreamsConfig.VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
+        streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, SCHEMA_REGISTRY_URL);
+
+        KStream<String, GenericRecord> locationsGeneric = builder.stream(INPUT_TOPIC);
+        locationsGeneric.print();
+        KStream<Long, Location> locations = locationsGeneric.map((k, v) -> new KeyValue<Long, Location>((Long) v.get("id"), (Location) new Location ((Long) v.get("id"), (String) v.get("name").toString(), (Long) v.get("sale")) ));
```
dguy: You don't need the cast `(Location)`

kafka-connect-streams/src/main/java/io/confluent/examples/connectandstreams/jdbcgenericavro/StreamsIngest.java
```diff
@@ -0,0 +1,97 @@
...
+        locations.print();
+
+        KStream<Long,Long> sales = locations.map((k, v) -> new KeyValue<Long, Long>(k, (Long) v.get("sale")));
```
dguy: as above

PR #131, opened 2017-06-28
Fix failing FanoutLambdaIntegrationTest

bbejeck:
ping @guozhangwang @mjsax @enothereska @dguy @miguno  for review


mjsax:
@bbejeck Why did it fail? This config should be backward compatible?

enothereska:
I couldn't repro locally btw, but the config change is probably good. It is strange that it's not backwards compatible though, I agree.

bbejeck:
I can reproduce locally, and I agree with @enothereska that it's good to update the config.

PR #137, opened 2017-07-22
reset pom files to use SNAPSHOT

bbejeck:


bbejeck:
@guozhangwang @dguy merging fix for pom files.

PR #161, opened 2018-05-28
Fixing build failure on "mvn clean package"

SimonLeigh:
Fixing build failure (mvn clean package) due to SNAPSHOT tag on versions of Kafka and Confluent. Currently our examples page doesn't work as a result: https://docs.confluent.io/current/quickstart/cloud-quickstart.html#step-5-run-java-examples

### DIFF
--- a/ccloud/java-clients/pom.xml
+++ b/ccloud/java-clients/pom.xml
@@ -35,9 +35,9 @@
 
     <properties>
         <java.version>1.8</java.version>
-        <kafka.version>1.1.0-SNAPSHOT</kafka.version>
+        <kafka.version>1.1.0</kafka.version>
         <kafka.scala.version>2.11</kafka.scala.version>
-        <confluent.version>4.1.0-SNAPSHOT</confluent.version>
+        <confluent.version>4.1.0</confluent.version>
         <slf4j-api.version>1.7.6</slf4j-api.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>


ewencp:
@SimonLeigh Snapshots are kind of unfortunate but a java-ism. The way this works is that the development branches should have -SNAPSHOT versions, which allows us to continue testing against the latest and greatest. When we actually tag a release, we *should* be generating a corresponding tag here where we also update the dependency versions to depend on the specific release. It looks like we haven't been creating the tags, though there are some -post branches that seem to have updated non-snapshot versions.

There are 2 potential follow ups here. First, we should probably get this repo into our normal release process so it gets the appropriate tags & branches. Second, if it makes sense to handle this repository differently, we can, but we should make sure we actually have a process in place for how to handle it.

ybyzek:
@SimonLeigh : there's another `SNAPSHOT` reference in line 7 that needs to be updated (not sure why I can't comment on the line itself...)

@ewencp : would it make sense to let this PR merge (with additional change mentioned above) and then we put it on the same branch strategy as a separate PR?

ewencp:
@ybyzek Sure, we're just going to have to make sure to update some of this manually until those changes are in. From https://github.com/confluentinc/common-tools/blob/master/confluent/config/dev/projects.json#L183 it looks like this is handled manually. It should probably match the group that kafka-streams-examples is in (which would tie it in with normal releases). However, the kafka-streams-example repo must also have some manual management going on or something because they also address this by still having SNAPSHOT, but making the default branch for GitHub a -post branch (which have final release values in them).

PR #165, opened 2018-09-18
DEVX-304: add link from examples to quickstart-demos

ybyzek:


README.md
```diff
@@ -1,14 +1,17 @@
-# Confluent Examples
+# Update Your Links
```
joel-hamill: should this instead say something like `Content has moved?`

